PR #3, opened 2026-06-09
Auto-load ext for crash-dump !analyze; add 0x9F walkthrough + sample

glslang:
## Why

The crash-dump playbook's first real step — `!analyze -v` — silently returned **empty** ("No export analyze found"). Root cause: the bundled WinDbg engine ships **no debugger extensions** (`winext\ext.dll` etc.), so no `!`-command resolves. The `ttd\` engine bits are bundled but `winext\` was not.

## Changes

- **`src/server.rs`** — `open_dump` now runs a best-effort `.load ext` after the dump settles, so `!`-extension commands resolve. No-op (open still succeeds) when `winext\` isn't bundled.
- **README + skill** (`setup.md`, `crash-dump.md`, `SKILL.md`) — bundle the WinDbg `winext\` extensions alongside `ttd\`, and use the **module-qualified `!ext.analyze -v`** (this minimal engine doesn't resolve bare `!analyze` even after `.load ext`). Also document:
  - naming a driver by hand (device-stack walk) when `!ext.analyze` leaves `MODULE_NAME: Unknown_Module`, and
  - the partial-minidump `0x80040205` read gotcha (uncaptured pages raise an engine exception, not a clean read error).
- **`docs/crash-dump-walkthrough.md` + `docs/samples/052126-34312-01.dmp`** — a worked example: a real `0x9F DRIVER_POWER_STATE_FAILURE` kernel minidump traced to `nvlddmkm.sys` (NVIDIA) via `!ext.analyze -v` and a manual `_DEVICE_OBJECT`/`_DRIVER_OBJECT`/`_IRP` walk.

## Verification

`cargo build --release` clean. Driving the rebuilt server as a sequential MCP client (`open_dump` → `execute {!ext.analyze -v}`): `.chain` confirms `winext\ext.dll`/`kext.dll` loaded by `open_dump`, and `!ext.analyze -v` returns the full triage (`DRIVER_POWER_STATE_FAILURE (9f)`, `FAILURE_BUCKET_ID: 0x9F_3`) with **no manual `.load`**.

## Notes

- `docs/samples/052126-34312-01.dmp` is a 5.8 MB binary committed as the walkthrough's sample.
- Out of scope (intentionally no fix): under *pipelined* concurrent requests the server dispatches tool calls in nondeterministic order (a call can race ahead of `open_dump`). Standard MCP clients await each call→result, so it doesn't occur in practice, and forcing sequential dispatch would block cancellation during a long `go`. Judged not worth a code change.

[...]

PR #4, opened 2026-06-09
Document the "one command at a time" concurrency contract

glslang:
## Why

Follow-up to #3. While verifying that PR I observed that the server dispatches **concurrent in-flight requests in nondeterministic order** — a pipelined tool call can race ahead of `open_dump` and fail with `0x80040205`. The engine serializes *execution* (single mpsc consumer on one thread), but not *submission order*.

This is **not worth a code fix**: standard MCP clients await each call→result before sending the next, so it never occurs in practice; debugger commands are stateful/order-dependent, so ordering guarantees wouldn't make pipelining safe anyway; and forcing sequential dispatch in rmcp would block `end_session`/cancellation from arriving during a long-running `go`. So this PR just **documents the contract**.

## Changes (docs only)

- **README** "Limitations & notes" — expand the single-session bullet into "one debug session, one command at a time": issue calls sequentially, await each result; concurrent pipelining can race `open_dump` (`0x80040205`) and is unsafe for stateful commands.
- **skill `SKILL.md`** cross-cutting gotchas — the same contract, concisely.

Both note that this only affects custom/batched callers, not standard MCP clients.

## Documentation

* Enhanced debugging session guidance to clarify command sequencing requirements for reliable operations.
* Expanded instructions on ending active sessions before opening new debugging targets.
* Clarified that debug commands must be issued sequentially to prevent operational failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,14 @@ timeline). For anything else, `dx` evaluates arbitrary data-model/LINQ expressio
   (after a `go`/breakpoint, not straight off a `!tt`) so the module's PDB is matched and loaded.
   With those, e.g. `ttd_calls("ucrtbase!__stdio_common_vfprintf")` returns the exact call count.
   Without symbols, the data model, navigation, and memory reads still work — query by address.
-- One debug session at a time (single engine instance).
+- **One debug session, one command at a time.** A single engine instance runs on a dedicated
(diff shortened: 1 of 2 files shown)


glslang:
@codex review

glslang:
@cursor review

PR #21, opened 2026-06-17
feat: driver IOCTL discovery & user-mode reachability

glslang:
## What

Adds a **static-first, dynamic-confirm** workflow for enumerating a driver's IOCTL surface and deciding, per caller token, whether each code is actually reachable from user mode. Motivated by using the live-kernel capability to debug a **driver**.

The framing "breakpoint and verify which IOCTLs are available and whether they're reachable" is reframed honestly:
- **Breakpoints confirm; they don't enumerate** — the available set is the static `switch (IoControlCode)` in the dispatch routine (recovered with `uf`/Binary Ninja); the bp sweep only confirms candidates and observes runtime behavior.
- **"Reachable" is a four-layer gate, per token:** *openable* (device DACL) → *namespace-visible* (symbolic-link scope) → *deliverable* (IOCTL `RequiredAccess` checked by the I/O manager before the IRP reaches the driver) → *handled vs. rejected* (`default:` → `STATUS_INVALID_DEVICE_REQUEST`).

## New MCP tools (`src/server.rs`)

| Tool | Wraps | Purpose |
|------|-------|---------|
| `decode_ioctl` | *(pure)* | Decode a 32-bit `CTL_CODE`; flags `METHOD_NEITHER` / `FILE_ANY_ACCESS` |
| `driver_object` | `!drvobj <name> 7` | Dispatch table — `MajorFunction[0x0e]` is the IOCTL handler |
| `device_object` | `!devobj` | Device type / characteristics / `SecurityDescriptor` (openable gate) |
| `irp_stack` | `!irp` | Current `IO_STACK_LOCATION`; IRP defaults to `@rdx` at a dispatch break |
| `ioctl_trace` | logging `bp` + `gc` | Prints each `IoControlCode` + buffer lengths for the sweep |

## Docs & examples

- **New playbook** `skills/windbg-debugging/driver-ioctl.md` — the gate model, WinDbg-native `uf` enumeration (default), Binary Ninja escalation with a JSON IOCTL-map handoff schema, the dynamic-confirm sweep, and a pitfalls/constraints box (local kernel = read-only; no-PDB → rebase).
- Surfaced in `SKILL.md`, `README.md`, `CHANGELOG.md` (`[Unreleased]`, no version bump).
- `examples/sweep_ioctls.ps1` (host-side KDNET driver) + `examples/send_ioctls_target.ps1` (target-side `DeviceIoControl` sender, run low-priv then elevated).

## Verification

- `cargo test` — **16 pass**, including two new `decode_ioctl_text` cases (`IOCTL_DISK_GET_DRIVE_GEOMETRY`; a `METHOD_NEITHER` + `FILE_WRITE_DATA` case).
- `cargo fmt --check` and `cargo clippy --all-targets` — clean.
- `claude plugin validate . --strict` — passed.
- Both PowerShell scripts parse; the target sender's embedded C# (`Add-Type`) compiles and fails cleanly on a missing device.

> **Note:** the optimized `cargo build --release` was not run locally because the running `windbg` MCP server holds `target/release/windbg-mcp.exe` locked (compilation succeeds; only the final file-replace is blocked). The identical code compiles clean under the dev profile + clippy, and CI's `release.yml` builds release on a fresh runner.

## Summary by CodeRabbit

* **New Features**
  * Added new MCP tools to decode IOCTLs, inspect driver/device details, view IRP stack information, and perform breakpoint-based IOCTL tracing.
  * Added example PowerShell scripts to sweep likely IOCTLs from a host and probe user-mode reachability from a target VM.

* **Documentation**
  * Added a new driver IOCTL playbook covering static-first enumeration with dynamic confirmation using an openable → namespace-visible → deliverable → handled reachability model.
  * Updated the changelog, Tools table, and example documentation, plus setup guidance for required debugger extensions.

* **Tests**
  * Added unit coverage for IOCTL decoding output and warning rendering.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Driver IOCTL discovery & user-mode reachability.** A new
+  [`driver-ioctl.md`](skills/windbg-debugging/driver-ioctl.md) playbook documents a
+  static-first, dynamic-confirm workflow for enumerating a driver's IOCTL surface and
+  testing whether each code is reachable from user mode (the openable → namespace →
+  deliverable → handled gate model), with WinDbg-native (`uf`) static enumeration by
+  default and Binary Ninja as an optional escalation. Five supporting tools:
+  - `decode_ioctl` — decode a 32-bit control code into its `CTL_CODE` fields and flag
+    `METHOD_NEITHER` / `FILE_ANY_ACCESS` (pure; no session needed).
+  - `driver_object` — dump a driver's dispatch table + devices (`!drvobj <name> 7`).
+  - `device_object` — inspect a device object's type/characteristics/SecurityDescriptor
+    (`!devobj`) to answer the *openable* gate.
+  - `irp_stack` — dump an IRP's current `IO_STACK_LOCATION` (`!irp`), defaulting the IRP
+    to `@rdx` at a dispatch break.
+  - `ioctl_trace` — install a conditional logging breakpoint at the IOCTL dispatch
+    routine that prints each `IoControlCode` + buffer lengths and continues.
+  - An `examples/sweep_ioctls.ps1` harness driving the dynamic confirm sweep.
+
 ## [0.1.3] - 2026-06-14
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     to `@rdx` at a dispatch break.
   - `ioctl_trace` — install a conditional logging breakpoint at the IOCTL dispatch
     routine that prints each `IoControlCode` + buffer lengths and continues.
-  - An `examples/sweep_ioctls.ps1` harness driving the dynamic confirm sweep.
+  - An `examples/sweep_ioctls.ps1` harness (host) + `examples/send_ioctls_target.ps1`
+    (target-side `DeviceIoControl` sender) driving the dynamic confirm sweep.
+  - `attach_kernel` / `attach_kernel_local` now `.load kdexts` automatically so the
+    `!drvobj`/`!devobj`/`!irp` commands behind `driver_object`/`device_object`/`irp_stack`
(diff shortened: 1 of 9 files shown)


glslang:
Addressed the review feedback in b323034 (all findings verified against current code first):

**`src/server.rs`**
- `decode_ioctl` now rejects codes wider than 32 bits via `u32::try_from` instead of silently truncating; `decode_ioctl_text` takes `u32`. (CodeRabbit)

**`examples/send_ioctls_target.ps1`**
- `-Codes` is now `[string[]]` parsed through a `ConvertTo-UInt32` helper, so the full unsigned 32-bit space works (e.g. `0x80008003`, device type with bit 31 set) — verified it runs. (Codex + CodeRabbit)
- Added a `-DesiredAccess` parameter (default `0` = minimal, so the `FILE_ANY_ACCESS`/minimal-handle case is testable; pass `0x80000000`/`0x40000000`/`0xC0000000`). (Codex)
- Open with `FILE_SHARE_READ_WRITE`, and the failure message now distinguishes access-denied (gle=5) from a sharing violation (gle=32). (CodeRabbit)

**`examples/sweep_ioctls.ps1`**
- Replaced `-RunMs` (only ever the JSON-RPC client timeout) with a `-Sweeps` loop of `go` windows. The server caps each `go` at `EXEC_WAIT_MS` (~60s) regardless of the client timeout, so looping is how the collection spans the operator running the sender as two tokens. (Codex)

**`skills/windbg-debugging/driver-ioctl.md`**
- Fixed MD038 (trailing space in code span) and MD040 (untyped fence) — `markdownlint-cli2` is now clean locally; removed the stray `</content>` marker.
- Per the request that the BN path can pin `IO_STACK_LOCATION` exactly: added a step noting that with NT types applied, Binary Ninja gives the precise `_IRP`/`_IO_STACK_LOCATION` offsets (and the dispatch routine dereferences them directly), confirming the `poi(@rdx+0xb8)+…` chain `ioctl_trace` uses rather than relying on runtime `dt`.

**`skills/windbg-debugging/setup.md`**
- Documented that the running `windbg` MCP server locks `target\release\windbg-mcp.exe`, so it must be disconnected before a from-source release build / end-to-end test; use the dev profile (`cargo test`/`clippy`) to iterate otherwise.

Local verification: `cargo test` (16 pass), `cargo clippy --all-targets`, `cargo fmt --check`, `markdownlint-cli2`, and `claude plugin validate . --strict` all clean.

PR #40, opened 2026-07-19
fix(ttd): repair index_trace, add record_trace env/cwd, clearer messaging

glslang:
## What

Fixes and small improvements to the TTD tools, all surfaced while using them to solve an obfuscated crackme (companion to the walkthrough in #39).

## Fixes

- **`index_trace` was broken.** It ran `!tt.index`, which fails with `LoadLibrary(tt) failed, Win32 error 0n2` — there is no `tt` extension. The bundled engine provides TTD indexing through `TtdExt.dll`, so it now runs **`!ttdext.index`**, building a persistent `.idx` next to the `.run` (verified against a real trace: 73 keyframes in ~1.6 s).
- **`open_trace` now flags an unindexed trace.** A freshly recorded `.run` has no `.idx`, so the first data-model query silently builds an in-memory index and can run long. `open_trace` now reports this (via the read-only `!ttdext.index -status`) and points at `index_trace`.
- **`registers` no longer returns a blank result** when there's no thread context (a module-load break or a bare `goto_position 0`) — it explains why and how to get a settled context.

## Added

- **`record_trace` `env` + `working_dir` options.** Pass `"env": ["KEY=VALUE", …]` and/or `"working_dir"`, applied to the recorded target — for programs that won't run without a specific environment (a Qt app's `QT_QPA_PLATFORM_PLUGIN_PATH`, an anti-analysis "run me from here" guard). Previously the recorder only inherited the server's environment, so such targets couldn't be recorded through `record_trace` at all.
- **Skill `ttd.md`** — the corrected index/record guidance, the **unbounded-search-wedges-the-engine** pitfall + recovery, and a "tabulate a pure function for a solver" technique.

## Not included (offered as a follow-up)

The deeper wedge fix — *interrupting* a runaway engine command (an unbounded `s` scan pins the single engine thread, so every later call times out behind it) — belongs one layer down in `win-kexp`'s `execute_command`. It already has the `InterruptHandle`/`SetInterrupt` watchdog used by `wait_for_event_bounded`; that pattern just needs applying around `Execute`. Happy to send it as a companion `win-kexp` PR + a pin bump here.

## Testing

- `cargo fmt --all --check`, `cargo clippy --all-targets`, `cargo test` — all clean (37 tests, +3 for `KEY=VALUE` parsing).
- `markdownlint-cli2` clean on the changed docs.
- Verified `!ttdext.index` and `!ttdext.index -status` against a real `.run` with the bundled engine.

## Summary by CodeRabbit

* **New Features**
  * Trace recording can now pass extra environment variables and set an optional working directory.
  * Added guidance for evaluating pure functions directly on a trace.
* **Bug Fixes**
  * Improved pre-flight detection and messaging for unindexed traces, including guidance to index first.
  * Updated trace indexing/opening to use the correct tooling and build/repair persistent indexes.
  * Register queries now explain when live thread context isn’t available.
* **Documentation**
  * Refreshed “Get a trace” and troubleshooting guidance, including cautions for memory-intensive searches.
* **Tests**
  * Added coverage for strict `KEY=VALUE` environment entry parsing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - [`docs/driver-ioctl-walkthrough.md`](docs/driver-ioctl-walkthrough.md): a worked
     `\Driver\mountmgr` enumeration + reachability report against a live kernel. The
     playbook now ends with a "Write the report" step + template.
+- **`record_trace` `env` and `working_dir` options** — pass extra `KEY=VALUE` environment
+  entries and a working directory to the recorded target, for programs that refuse to run
+  without a specific environment (e.g. a Qt app's `QT_QPA_PLATFORM_PLUGIN_PATH`, or an
+  anti-analysis "run me from here" guard). Previously the recorder only inherited the
+  server's environment.
+
+### Fixed
+
+- **`index_trace` now works.** It invoked `!tt.index`, which fails with `LoadLibrary(tt)` —
+  there is no `tt` extension. The bundled engine exposes trace indexing through `TtdExt.dll`,
+  so `index_trace` now runs `!ttdext.index` (building a persistent `.idx` next to the `.run`).
+- **`open_trace` flags an unindexed trace.** A freshly recorded `.run` has no `.idx`, so the
+  first data-model query silently builds an in-memory index and can run long; `open_trace`
+  now says so up front (via `!ttdext.index -status`) and points at `index_trace`.
+- **`registers` no longer returns a blank result** when there is no thread context (a
+  module-load break or a bare `goto_position 0`); it explains why and how to get a context.
 
 ## [0.1.3] - 2026-06-14
 
--- a/skills/windbg-debugging/ttd.md
+++ b/skills/windbg-debugging/ttd.md
@@ -13,11 +13,18 @@ TTD positions are `major:minor` (a sequencing point and a step within it), not w
 ## 1. Get a trace
 
 - **Record** (Administrator, `TTD.exe` on `PATH`):
-  `record_trace { "out_dir": "C:\\traces", "target": "C:\\path\\app.exe arg" }`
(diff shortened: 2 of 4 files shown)


glslang:
Addressed the review in 48e8147:

- **open_trace ordering** — now runs `!ttdext.index -status` (reads the on-disk `.idx` only, never builds) *before* the `dx @$curprocess.TTD.Lifetime` query, so the index-state check never sits behind a data-model query.
- **out_dir vs working_dir** — `record_launch` now resolves `out_dir` to an absolute path (`std::path::absolute`), so a relative `out_dir` combined with `working_dir` no longer mismatches between where the dir/log are created and where TTD.exe writes `-out`.
- **env validation** — `env` entries are parsed/validated before the log file is created, so a malformed `KEY=VALUE` no longer leaves a stray empty log behind.

`cargo fmt`/`clippy`/`test` clean (37 tests). The engine-level wedge fix is going out as the companion `win-kexp` PR now.

glslang:
Companion engine-level fix (the wedge): glslang/win-kexp#68 — adds `execute_command_bounded` so a runaway command self-interrupts instead of pinning the engine thread. Once it merges I'll do the pin bump + route the `execute`/`dx` paths through it here.

glslang:
Addressed the index-status review in 8e31d64: the predicate now warns unless `!ttdext.index -status` is exactly `Index file loaded.` — so a stale/out-of-date/corrupt/unloadable `.idx` also warns (per the MS TTD guidance you linked), not just the missing case. A blank status (query failed) no longer produces a false warning. `cargo fmt`/`clippy`/`test` clean (37 tests).

glslang:
Addressed in 58410c3: `index_trace` now runs `!ttdext.index -force`, which deletes and rebuilds an unloadable/corrupt `.idx` (plain `!ttdext.index` can't repair those) while staying a fast no-op on a healthy index (verified: 2ms no-op on an already-loaded index). So following the `open_trace` "index not loaded" note now actually repairs those traces.

PR #45, opened 2026-07-20
fix: bound execute/dx/ttd_* commands so a runaway search can't wedge the engine

glslang:
## What

Closes the engine-wedge end to end — the companion adoption of the now-merged glslang/win-kexp#68. An unbounded debugger command (most importantly a broad `s` memory search) used to pin the single engine thread indefinitely, so every subsequent tool call timed out behind it and the only recovery was to kill and reconnect the server.

## Change

- **Pin bump.** `win-kexp` → the commit that adds `execute_command_bounded` (+ its stale-interrupt drain).
- **`EngineHandle::run_command`** (`engine.rs`) — runs a raw command bounded by an engine-side watchdog that `SetInterrupt`s ~15 s before the per-call timeout, so a runaway command self-aborts (with a note) and frees the thread rather than wedging the session. Quick, inherently-bounded ops keep using `run`.
- **Routed** `execute`, `dx`, `ttd_calls`, `ttd_memory`, `ttd_events` through `run_command` — the surface where a runaway `s`/`dx` actually happens.

The budget is `call_timeout − 15 s` (285 s), so nothing that would have completed within the server's own timeout is aborted; only commands that exceed it — i.e. the wedge case — get interrupted.

## Testing

- `cargo fmt --all --check`, `cargo clippy --all-targets`, `cargo test` — all clean (34 tests).
- Builds against the bumped pin.

## Notes

Independent of #40 (the TTD index/record/messaging fixes); both touch `src/server.rs` but different tools, so at most a trivial merge. The wedge pitfall stays documented in the `ttd.md` skill for the manual-recovery case (older pins).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented long-running debugger commands from permanently blocking the session.
  * Commands that exceed their execution limit now stop automatically and provide an explanatory message.
  * Improved reliability for `execute`, `dx`, and time-travel debugging commands, keeping subsequent tool calls responsive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **`registers` no longer returns a blank result** when there is no thread context (a
   module-load break or a bare `goto_position 0`); it explains why and how to get a context.
 
+### Fixed
+
+- **A runaway debugger command no longer wedges the session.** `execute`, `dx`, and the
+  `ttd_*` query tools now run through a bounded path
+  ([`win-kexp`](https://github.com/glslang/win-kexp)'s `execute_command_bounded`) that
+  `SetInterrupt`s the engine shortly before the per-call timeout. Previously an unbounded
+  command — most importantly a broad `s` memory search — could pin the single engine thread
+  indefinitely, so every later tool call timed out behind it and the only recovery was to
+  kill and reconnect the server. Now such a command self-aborts (with a note) and the engine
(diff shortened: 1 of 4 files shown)


glslang:
Addressed in the latest commit: `run_command` now budgets the watchdog against the *remaining* outer timeout, not the full `call_timeout`. It captures the submission instant and, on the engine thread, computes `call_timeout - queued_wait - 15s` (floored at 15s so a command dequeued near the deadline still arms the watchdog). So the Ctrl+Break fires ~15s before the caller gives up regardless of how long the command sat in the queue behind another job — closing the re-wedge you spotted. `fmt`/`clippy`/`test` clean.

PR #59, opened 2026-08-01
fix: commit the session handle for the attaching and launching openers

glslang:
> [!NOTE]
> **Unblocked.** [glslang/win-kexp#72](https://github.com/glslang/win-kexp/pull/72) is merged, and
> `Cargo.lock` is repinned from `6644f0f` to `47993c4` — without that, CI would resolve a win-kexp
> revision predating the `*_begin` openers and fail on four missing methods. No git rebase was
> needed: this branch is 0 commits behind `main`.

## The gap this closes

0.3.0 shipped per-opener session handles, but only `open_dump` and `open_trace` could actually
commit one. win-kexp fused the target-creating call with the wait for the initial break into a
single `Result`, so for the other four a failure could mean either:

- nothing happened — the slate is clean, re-open; or
- the target *was* created or attached, and only the wait failed — never re-open.

Indistinguishable from here, and needing opposite recovery. `OpenSideEffect` existed solely to
hedge on those four, pointing callers at `vertarget` instead of claiming a retry was safe. That
removed the bad advice without removing the ambiguity — the handle still wasn't committed,
because there was nowhere to commit it.

## The seam

win-kexp#71 splits the openers: `x_begin()` returns a `PendingTarget` guard that cannot exist
unless the side effect succeeded, and `wait()` completes the initial break.

`opened_result`'s `transition` now receives a `commit` callback to invoke there, so all six
openers read identically:

```rust
let pending = e.attach_process_begin(pid).map_err(es)?;
commit();
pending.wait().map_err(es)
```

A failure is then sorted by which side of the seam it fell on — before it, the error stands
alone and the open is recorded `Failed`; after it, the error carries the `session_id` and says
not to open again. Dumps and traces move to the same shape, their load wait crossing from
`report` into `transition`, so all six are uniform rather than two-of-a-kind.

`OpenSideEffect` and its caveat delete entirely.

## Notes for review

- **The defensive commit.** A `transition` that returns `Ok` without committing would leave the
  caller with an open target and no handle. None do; it commits anyway rather than let a future
  opener silently regress the guarantee.
- **`post_commit_failure` is a free function** for the same reason `failure_caveat` was one — so
  the message tests without an engine. The old test is replaced by one over it.
- **The queued-precheck design is untouched.** The gate still runs in the same queued job, so
  the ordering guarantee in the CHANGELOG holds unchanged.

## Validation

| | |
|---|---|
| unit tests | 78 pass |
| clippy | clean (`--all-targets`) |
| smoke, debugger tier | `WINDBG_MCP_SMOKE_DUMP=1` — `a_dump_session_opens_reads_and_closes` passes against the sample dump, exercising the restructured `open_dump` path end to end |

Re-verified against the **merged** win-kexp `main` (`47993c4`), not a local path patch — build,
clippy, tests and smoke all run against the same revision `Cargo.lock` names.

**Not validated:** the two kernel attaches, for want of a KDNET target. Tracked as
[glslang/win-kexp#73](https://github.com/glslang/win-kexp/issues/73) and added to `FOLLOWUPS.md`
as item 12 so it does not rest in a PR body.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session identifiers are now retained and returned even when initial target connection or break-in waits fail.
  * Improved recovery guidance is provided when opening a dump, trace, kernel, process, or launched target encounters a post-connection error.
  * Target reopening now clearly distinguishes safe retries from targets that have already been created.

* **Documentation**
  * Updated the unreleased changelog and follow-up documentation with details on connection handling and validation work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Every opener now commits its session handle, including the four that attach or launch.**
+  0.3.0 shipped this guarantee for `open_dump` and `open_trace` only, and documented the gap
+  for the rest: win-kexp fused the target-creating call and the wait for the initial break
+  into one `Result`, so a failure could mean "nothing happened" or "the process started /
+  the attach succeeded, then the wait failed" — indistinguishable from here, and needing
+  opposite recovery. Those four tools hedged accordingly, telling callers to check
+  `vertarget` before opening again instead of claiming a retry was safe.
+
+  win-kexp split them (glslang/win-kexp#71): each opener is now a `x_begin()` returning a
+  `PendingTarget` guard, plus a `wait()` on that guard. The guard cannot exist unless the
+  side effect succeeded, so there is finally a seam to commit at. `opened_result` hands its
+  `transition` a `commit` callback to invoke at that seam, and every opener reads the same
+  way — side effect, `commit()`, wait.
+
+  So a failed break-in wait on `attach_process`, `attach_kernel`, `attach_kernel_local` or
+  `launch` now returns the error *with* a usable `session_id`, exactly as a failed load wait
+  on a dump already did. The hedge is gone: the server knows which side of the seam a failure
+  fell on, and says so — re-open when nothing was created, never re-open when the target is
+  already there. For `launch` that is the difference between one process and two.
+
 ## [0.3.0] - 2026-08-01
 
 ### Added
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -843,7 +843,7 @@ dependencies = [
 [[package]]
 name = "win-kexp"
 version = "0.1.0"
-source = "git+https://github.com/glslang/win-kexp?branch=main#6644f0f2224544deea01b11c46bcf1e14979f696"
+source = "git+https://github.com/glslang/win-kexp?branch=main#47993c41e8d5942a5efec6fc2452538c064951c5"
 dependencies = [
  "byte-strings",
  "goblin",
--- a/FOLLOWUPS.md
+++ b/FOLLOWUPS.md
@@ -1,8 +1,9 @@
 # Follow-ups
 
-Deferred work, in two clusters: items 1–6 come from the reachability-confirmation effort (path
+Deferred work, in three clusters: items 1–6 come from the reachability-confirmation effort (path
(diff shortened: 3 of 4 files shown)


glslang:
Both comments were against 1c63709, before the repin.

**Codex (P1, lockfile).** Already fixed in b0068c9 — independent confirmation of the same problem, and a correction to what this PR's description originally claimed. `branch = "main"` in `Cargo.toml` governs *resolution*, but the lockfile pins an exact revision, so tracking the branch never refreshed it. Repinned `6644f0f` → `47993c4` and re-verified against the merged win-kexp `main` rather than a local path patch.

**CodeRabbit (wording).** Taken in 9045b26. Agreed this is one of the few strings here where phrasing has functional rather than cosmetic weight — a calling agent reads it to decide whether to retry an open, and "for launch/attach that starts a second process or attaches twice" leaves "that" pointing at nothing obvious. Now: "for launch or attach, doing so would start a second process or attach a second time."

Also tightened the test that covers it. It asserted on either half of the phrasing with an `||`, so it would have stayed green if one of the two costs went missing; it now requires both, since the point is that the advice names the cost for launch and attach alike.

glslang:
Checked this against a live engine rather than reasoning about it, and the premise does not hold — no behaviour change needed. Evidence recorded as a comment in b5ad247.

The concern is that `CreateProcessWide` is deferred, so a launch that creates nothing would fail inside `pending.wait()`, after `commit()`. But only the *spawn* is deferred; validation is not. `CreateProcessWide` resolves and checks the image synchronously and fails on the spot:

| command line | result |
|---|---|
| missing path | `begin=Err` 0x80070002 — cannot find the file |
| a directory | `begin=Err` 0x80070005 — access denied |
| non-PE file | `begin=Err` 0x800700C1 — not a valid Win32 application |
| empty string | `begin=Err` 0x80070057 — parameter incorrect |

All four return on the `?` in `launch_process_begin`, before `commit()` runs, so they are reported as clean-slate failures with no handle committed and no warning against retrying. `execution_status` afterwards is 7 (`DEBUG_STATUS_NO_DEBUGGEE`), confirming nothing was created. That covers both cases named — "the executable is missing or access is denied".

Worth noting the inverse, since it is what the design rests on: because validation is eager, a failure from `wait` means the process really *was* created. That is what makes the post-commit message ("a session exists, do not open again") true rather than a guess.

I have left the finding in the code as a comment rather than just closing this out — it is a reasonable question to ask of that seam, and the answer is not visible from the source, since it lives in what `CreateProcessWide` validates eagerly versus what it defers.

PR #72, opened 2026-08-04
Give the worker protocol a channel nothing else can write on

### LINKED ISSUE #65: Worker protocol shares stdout with anything the engine prints
Found by Codex reviewing #62 (`src/engine.rs`, `next_message`). Mitigated there in a5fc0e6; this is the fuller answer.

## The exposure

The supervisor↔worker protocol is line-delimited JSON on the worker's **stdout**, the same handle any code in that process can write to. DbgEng's own output does not come this way — it is captured through `IDebugOutputCallbacks` — but an extension DLL that writes to the console directly does.

`next_message` skips a line it cannot parse. That is right for a stray line, and wrong when the stray text is *unterminated*: the next protocol message is appended to it, the combined line does not parse, and the reply is discarded. The failure is worse than losing one message, because only receipt of `Done` removes a waiter — so the caller times out, the waiter stays, and `busy()` keeps that session unreclaimable for the life of the server.

## What a5fc0e6 does, and does not do

Each message now opens with a newline as well as ending with one, so an unterminated line before it is terminated and the message always starts its own line. That removes the failure above.

It is still a mitigation. The channel remains shared, so anything that manages to interleave *within* a write, or that writes after our leading newline and before our JSON on the same line, can still corrupt a message. Nothing in the current worker does either — only the engine thread emits protocol messages, and it emits them between ops — but the property is by construction unenforced.

## The fuller fix

Give the protocol a channel that extension output cannot reach:

- an extra inherited pipe handle passed to the child, with stdout left as a pure log sink; or
- a named pipe per worker, which also drops the line-oriented framing in favour of length prefixes.

Either makes "stray output cannot corrupt a reply" a structural property rather than a convention. Both need a little Win32 (`windows-sys` is already a transitive dependency, not yet a direct one), which is why it is not a drive-by change.

## Worth knowing when picking it up

The reason the current design tolerated a shared channel is that a corrupted line was assumed to cost one message. It does not — it costs the session, permanently. Whatever replaces it, keep the property that a lost `Done` cannot strand a waiter: settling a session on worker exit, or timing waiters out server-side, would bound the damage independently of the framing.

glslang:
Closes #65.

## The exposure

The supervisor↔worker protocol rode the worker's stdin and stdout — the same stdout any code in that process can write to. DbgEng's own output never lands there (it is captured through `IDebugOutputCallbacks`), but an extension DLL that prints to the console directly does, and an **unterminated** stray line swallowed the message written after it.

The supervisor drops what it cannot parse, and only a `Done` removes a waiter, so the cost of one stray `printf` was not a lost line but a lost session: the call timed out, its waiter stayed, and the session counted as busy — and so could never be reclaimed — for the life of the server. a5fc0e6 mitigated it by opening each message with a newline; the property was still a convention about who prints where.

## The channel

Each worker now gets a pair of anonymous pipes, created by the supervisor and inherited across the spawn: requests down one, messages up the other. An anonymous pipe has no name to open and is reachable only through an inherited handle, so nothing outside that pair of processes can write on it — "stray output cannot corrupt a reply" is a property of the plumbing rather than of what happens to be loaded.

Marking, spawning and closing all happen under one lock. A handle is inheritable from the moment it is marked until it is closed, and **every** process created in between gets a copy: two overlapping opens would cross-inherit, and a worker holding another worker's *write* end keeps that pipe from ever reporting EOF — so the supervisor would never learn the other worker had exited, and the calls it owed replies to would wait for ever. std holds a lock of its own across the same window for the stdio handles it prepares; this one covers the handles std knows nothing about.

Both directions moved, not just the one the report names. The request side carries the KD connection string, key and all, and a request consumed by something else costs the same session the same way — a reply that never comes.

## What the worker's standard handles are now

- **stdout**: piped and drained into the log, per session. Nothing of ours writes there, so anything that arrives was printed by something else in that process — which is the first time an extension's console output has been visible at all. Drained rather than discarded because an unread pipe fills and the *next* write blocks the engine thread inside DbgEng; read as bytes and decoded lossily, because an extension writing in the console's code page is not UTF-8 and a decode error must not be able to stop the drain.
- **stdin**: `NUL`, emphatically not inherited — the supervisor's stdin is the MCP transport, and a worker holding it could consume the client's requests.
- **stderr**: inherited, as before.

Anonymous pipes cannot be read or written asynchronously on Windows (they are not overlapped handles), so the message reader and the request pump are threads. `reader` stays a task: settling a session touches the registry and the worker's process handle, which is where both belong.

## Keeping a lost `Done` from stranding a waiter

The issue asks for this to be bounded independently of the framing, so it is answered directly:

- a result that cannot be encoded is replaced by a `Done` that says so — the caller loses the output, not the session;
- a channel that will not take a write means the supervisor is gone, its request channel is at EOF or about to be, and its exit fails every outstanding call out;
- worker exit still settles the session and fails its callers out, unchanged.

Teardown is otherwise unchanged in substance: EOF on the request channel now means what EOF on stdin meant, and a worker still releases its target before it exits. Comments, README, CLAUDE.md, `docs/smoke-test.md` and `examples/ctrl_c_teardown.ps1` say so rather than continuing to claim stdin.

`windows-sys` becomes a direct dependency (`Win32_Foundation`, for `SetHandleInformation`). It was already in the tree via tokio, so nothing new is compiled.

## Verification

- `cargo test` — 104 unit + 18 smoke, all pass. New: the channel-handle parser (including "half a channel is no channel" and a null handle), log clipping, and `a_workers_stdout_is_not_its_protocol_channel`, where `cmd.exe` stands in for the hazard — handed a worker's command line it prints its banner and an unterminated prompt to stdout, says nothing on the channel, and exits. The supervisor must hear silence and then EOF, which also catches a supervisor copy of a child's end left open.
- Debugger tier (`$env:WINDBG_MCP_SMOKE_DUMP = "1"; cargo test --test mcp_smoke`) — all pass in ~20s: real workers handshake over the new channel, open the dump, coexist, a parked kernel attach is reclaimed by pid, and no worker outlives the connection.
- By hand against the dev build: a dump session opens, executes, reads registers and releases cleanly; pointing `WINDBG_MCP_WORKER_EXE` at `cmd.exe` shows its banner **and its unterminated prompt line** logged as `session …: its engine worker wrote to stdout: …`, with the open failing cleanly and the server unaffected.

## Summary by CodeRabbit

* **New Features**
  * Improved supervisor–worker communication using dedicated protocol channels.
  * Prevented extension output from interfering with protocol messages.
  * Added logging for stray worker output, including clipping of excessively long lines.
  * Improved worker shutdown when the request channel closes.

* **Documentation**
  * Updated protocol, lifecycle, troubleshooting and smoke-test documentation.
  * Added guidance covering channel failures and worker teardown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **The supervisor↔worker protocol has a channel of its own, so nothing a worker prints can cost a
+  session** ([#65](https://github.com/glslang/windbg-mcp/issues/65)).
+
+  The protocol used to ride the worker's stdin and stdout — the same stdout any code in that
+  process can write to. DbgEng's own output never lands there (it is captured through
+  `IDebugOutputCallbacks`), but an extension DLL that prints to the console directly does, and an
+  *unterminated* stray line swallowed the message written after it. The supervisor drops what it
+  cannot parse, and only a `Done` removes a waiter, so the cost of one stray `printf` was not a
+  lost line but a lost session: the call timed out, its waiter stayed, and the session counted as
+  busy — and so could never be reclaimed — for the life of the server. 0.4.1 mitigated this by
+  opening each message with a newline; the property was still a convention about who prints where.
+
+  Each worker now gets a pair of anonymous pipes, created by the supervisor and inherited across
+  the spawn: requests down one, messages up the other. An anonymous pipe has no name to open and
+  is reachable only through an inherited handle, so nothing outside that pair of processes can
+  write on it — "stray output cannot corrupt a reply" is a property of the plumbing rather than of
+  what happens to be loaded. The worker's stdout is now only a log: it is drained into the
+  server's stderr with the session it came from, which is also the first time an extension's
+  console output has been visible at all. Its stdin is `NUL` — never inherited, since the
+  supervisor's stdin is the MCP transport.
+
+  What the channel cannot make impossible is a `Done` that is never written, and that is the half
+  that strands a waiter, so it is answered directly: a result that cannot be encoded is replaced
+  by one that says so, and a channel that cannot be written to means the supervisor is gone — its
+  exit fails every outstanding call out. Teardown is unchanged in substance: EOF on the request
+  channel now means what EOF on stdin meant, and a worker still releases its target before it
+  exits.
+
 ## [0.4.1] - 2026-08-04
 
 ### Fixed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,7 +19,9 @@ worker supervision, routing), `src/worker.rs` (the child process and the engine 
 
 Practical consequences when debugging this server: a stack trace or log line can come from either
 role (worker logs are untagged by target and inherit the supervisor's stderr), and killing the
-supervisor leaves no workers behind — they exit when their stdin closes.
+supervisor leaves no workers behind — they exit when their request channel closes. That channel is
+a pair of inherited anonymous pipes, *not* the worker's stdio: anything a worker prints to stdout
+is drained into the log and cannot reach the protocol.
 
 ## Updating the running windbg MCP after code changes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -908,6 +908,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "win-kexp",
+ "windows-sys",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,6 @@ schemars = "1"
 anyhow = "1"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+# Only for the handle plumbing behind the supervisor↔worker channel (see `engine::spawn_worker`):
+# marking a pipe end inheritable is the one thing std cannot express.
+windows-sys = { version = "0.61", features = ["Win32_Foundation"] }
--- a/FOLLOWUPS.md
+++ b/FOLLOWUPS.md
@@ -1,11 +1,11 @@
 # Follow-ups
 
-Deferred work, in four clusters: items 1–6 come from the reachability-confirmation effort (path
+Deferred work, in five clusters: items 1–6 come from the reachability-confirmation effort (path
 recipe + `run_to_address`, merged 2026-07-04), items 7–11 from surveying this server against the
 MCP `2026-07-28` extensions (tasks, apps), item 12 from the opener split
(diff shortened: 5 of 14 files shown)


glslang:
Review addressed in a4e9845. Four of the five findings were real — including one the first commit's own doc comment claimed was impossible — so thanks to both reviewers.

**Codex P1 — `TTD.exe` is spawned outside the lock.** Correct, and the more serious half of it is that `SPAWN_LOCK`'s doc *asserted* nothing else in this process spawns children. Two things do: `ttd::record` launches the recorder, and two engine tests spawn a stand-in child. The inherit flag is a property of the process, not of the spawn that set it, so a `record_trace` landing inside the marking window hands `TTD.exe` a worker's message write end — and the recorder outlives the whole recording, so that pipe never reports EOF, `reader`'s tail never runs, and the session never settles.

Every `spawn()` in the crate now goes through `engine::spawn_guard()` (`src/engine.rs`), including `ttd.rs` and the tests, and the rule is written where the next spawn site will read it rather than left as an assertion about the present. A `STARTUPINFOEX` handle list would make it structural rather than disciplined, but that means leaving `std::process::Command` for a raw `CreateProcessW` — noted as the better long-term shape, not done here.

**CodeRabbit — the test stand-ins reach the same failure.** Same mechanism, and reachable under `cargo test`: the stand-ins outlive `a_workers_stdout_is_not_its_protocol_channel` by 30s, so one created inside the window would hold its channel open and the EOF assertion would fail on its 10s budget. Both now go through a `stand_in_child()` helper that takes the guard.

**CodeRabbit — bound the accumulation, not only the logged text.** Agreed, and it is a regression in kind rather than a nitpick: before this PR runaway output filled the worker's own pipe, which is self-limiting and costs one session; draining it eagerly moved that pressure into the supervisor's heap, where it costs every session. `next_capped_line` now caps what is **buffered** at `LOGGED_LINE_LIMIT` and discards the rest of an over-long line, and the log says how many bytes went. Test reads through an 8-byte buffer, so the cap has to hold across many small chunks rather than only within one.

**CodeRabbit — clear `HANDLE_FLAG_INHERIT` in the worker.** Done (`worker::stop_inheriting`), on both ends once the worker owns them. DbgEng creates children of its own — a launched debuggee, whatever an extension shells out to — and one that inherited the message end would keep the pipe open past the worker's exit, which is precisely the signal the supervisor settles a session on. Warned rather than fatal: the channel still works, and the risk only materializes if something is spawned there at all.

**CodeRabbit — smoke-test wording.** Fixed; the worker exits without a usable channel rather than going quiet.

Verification on a4e9845: 105 unit + 18 smoke tests pass, debugger tier (`WINDBG_MCP_SMOKE_DUMP=1`) passes in ~20s, and the by-hand probe — `WINDBG_MCP_WORKER_EXE` pointed at `cmd.exe` — still shows its banner *and its unterminated prompt line* drained into the log with the open failing cleanly and the server unaffected. CI is green on the new commit; CodeRabbit's re-review reports rate limited, so it has not re-checked it.

PR #73, opened 2026-08-06
pool: four MCP tools over win-kexp's pool walker

glslang:
> **Draft — blocked on [glslang/win-kexp#80](https://github.com/glslang/win-kexp/pull/80).** This needs win-kexp's `pool::query` API. Once #80 lands on `main` I'll `cargo update -p win-kexp` and push the lock bump, at which point CI can go green. Opening now so the two are reviewable together.

Adds `pool_find_tag`, `pool_chunk`, `pool_census` and `pool_diagnostics` (38 → 42 tools). All four go through win-kexp's own descriptor walk rather than `!pool`/`!poolused` text, reading one cached snapshot, so they cannot disagree with each other.

## Distinctions the tools keep deliberately

- **`pool_find_tag` reports only allocated chunks.** A freed chunk's tag is not reliably preserved by the allocator, so listing them would be inventing data. Ask about a specific possibly-freed address with `pool_chunk`.
- **`pool_chunk` separates "free hole in a walked region" from "not covered by the snapshot".** That is the difference between a dangling pointer and one that never pointed at pool; collapsing them would make the tool useless for the question it exists to answer. It also reports the neighbouring chunks — what a reclaim would land next to.

## Why `pool_diagnostics`

A real walk emits ~18k diagnostics across 135+ categories. Any per-call summary truncates, and the line explaining one specific heap is reliably *not* in the truncated head. A plain case-insensitive substring filter found a special-pool misclassification in a single call — and disproved a wrong hypothesis in another, which is the part that saved the most time.

Relatedly, every result now carries the walk's own state (chunks seen, diagnostics grouped by category) whenever it comes back empty. "The pool holds no such chunk" and "the walk reached almost none of the pool" rendering identically hid a real bug for three rounds.

## Notes for review

- `open_world_hint = true` on all four: layout resolution can trigger a PDB download, and over KDNET the target itself is remote. The existing invariant test caught me getting this wrong.
- Rendering lives in the worker; the typed data stays in win-kexp.
- `examples/messagemanager/` is a client for a CTF driver, run on the target VM — the server cannot issue `DeviceIoControl` itself.
- Golden `tools_list.json` re-recorded; the diff adds only the four tool names.

## Verification

114 unit + 18 smoke tests pass, `cargo fmt` and clippy clean. Exercised end to end against a live Windows 26100 KDNET target, including a driver under `verifier /flags 0x1` where `pool_find_tag Tgsm` resolves to `ffff8c8f13a02f90 / 0x68 / SpecialNonPagedNx`, matching `!pool` exactly, and `pool_chunk` shows the guard page as an `Unreadable` neighbour.

## Summary by CodeRabbit

* **New Features**
  * Added four read-only kernel pool inspection tools for tag searches, chunk inspection, diagnostics, and allocation censuses.
  * Added session-scoped snapshot caching, refresh controls, filtering, paging, coverage details, neighbouring-chunk information, and result limits.
  * Added a PowerShell client supporting MessageManager create, delete, data, flush, demo, stress, and race workflows.
* **Documentation**
  * Documented the tools, x64 kernel-target requirement, caching behaviour, filtering, and coverage distinctions.
* **Testing**
  * Added coverage for diagnostic grouping, filtering, address formats, and the expanded tool catalogue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -884,7 +884,7 @@ dependencies = [
 [[package]]
 name = "win-kexp"
 version = "0.1.0"
-source = "git+https://github.com/glslang/win-kexp?branch=main#38eeafd36cbd037a7fdda08b9209ad9bf60c1d5d"
+source = "git+https://github.com/glslang/win-kexp?branch=main#ce8bce2ad21bac007c835ef6ca198341737fae06"
 dependencies = [
  "byte-strings",
  "goblin",
--- a/README.md
+++ b/README.md
@@ -247,7 +247,7 @@ gh attestation verify <zip> --repo glslang/windbg-mcp `
 | TTD nav | `step_back` (`t-`), `step_over_back` (`p-`), `reverse_go` (`g-`), `goto_position` (`!tt`) |
 | TTD analysis | `ttd_calls`, `ttd_memory`, `ttd_events`, `index_trace`, `record_trace` |
 | Driver IOCTL | `decode_ioctl`, `driver_object`, `device_object`, `irp_stack`, `ioctl_trace`, `reachable_from_dispatch` |
-| Kernel pool | `pool_find_tag`, `pool_chunk`, `pool_census` |
+| Kernel pool | `pool_find_tag`, `pool_chunk`, `pool_census`, `pool_diagnostics` |
 | Raw     | `execute` — run any debugger command, returns full text output |
 
 ### Sessions and session handles
@@ -355,7 +355,7 @@ timeline). For anything else, `dx` evaluates arbitrary data-model/LINQ expressio
   exists, and the path is reported), while `NOT REACHABLE` is best-effort within the explored
   bounds. If the dispatch uses a `switch(IoControlCode)` jump table (common), pass the specific
   handler VA as `from` to scope past it, or confirm dynamically with a breakpoint + `go`.
-- The **kernel pool** tools (`pool_find_tag`, `pool_chunk`, `pool_census`) walk the allocator's own
+- The **kernel pool** tools (`pool_find_tag`, `pool_chunk`, `pool_census`, `pool_diagnostics`) walk the allocator's own
   descriptors through win-kexp rather than shelling out to `!pool`/`!poolused`, so all three read
   one snapshot and cannot disagree with each other. They need a **broken-in x64 kernel** target.
   Walking every pool page is expensive, so the snapshot is **cached per session** and reused; pass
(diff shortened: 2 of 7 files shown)


glslang:
Unblocked: win-kexp#80 is merged, so this now builds against `main` rather than a local path patch. `cargo update -p win-kexp` took the lock to `ce8bce2a`; **132 tests pass** (114 unit + 18 smoke), fmt and clippy clean.

Also surfaces `PoolSnapshotReport.complete`, which win-kexp gained during that PR's review. It is not implied by an empty diagnostics list — a walk can end partway through in silence — so the tools now print coverage explicitly instead of letting a truncated snapshot read as a healthy one. That distinction is the whole reason the field exists.

Marking ready for review.

Known follow-ups tracked against win-kexp, none blocking here: #82, #83, #85, #87.

glslang:
All five addressed in the latest commit — the P1 was worth catching.

**Unreadable spans reported as use-after-free** (Codex P1) — correct and the most important of these. `render_chunk` drew the UAF conclusion for any state other than `Allocated`, but `Unreadable` only means the walker could not read the span, and a Verifier guard page reads exactly that way. So the tool whose entire purpose is answering "is this pointer dangling" was turning a coverage limit into a false positive. Only `ReusableFree`/`CachedFree` now say use-after-free; `Unreadable` says the state is unknown and why.

**Unbounded limits** (CodeRabbit, Major) — correct. The worker builds each reply as one `String`, so an unbounded `limit` requests a snapshot-sized allocation: hundreds of thousands of chunks, or ~19k diagnostic lines on an idle machine. Clamped to 2000 rows.

**Empty diagnostics hid incompleteness** (CodeRabbit) — correct, and the same class of bug the `complete` flag was added for: "no diagnostics match" from a truncated walk read as a clean result. Now states coverage.

**README missing `pool_diagnostics`** (both reviewers) — fixed in the table and the limitations note.

132 tests, fmt and clippy clean.

glslang:
Second round addressed across `a86f71c`, `5630c09`, `a9daeb2`. Every finding was the same bug wearing a different hat, which is worth naming: **a result that says nothing about the walk's coverage gets read as a fact about the pool.** The first round fixed that for empty results; these fix it for the nonempty ones and for the prose describing them.

**Nonempty tag results hid partial coverage** (Codex P1) — correct, and the more dangerous half of the bug the first round fixed. `pool_find_tag` appended the walk report only when it found *nothing*. An empty answer at least looks like it needs explaining; "3 allocations, 0x138 bytes total" does not — it reads as the whole set whether the walk covered the pool or a corner of it. find_tag can only report what the walk indexed, so under partial coverage that count is a floor. New `coverage_caveat` helper says so, and says which way the number is wrong. Silent on a complete walk, or the warning stops meaning anything on the walk where it matters.

**Filtered diagnostics did the same** (Codex P2) — correct. "1 of 2 diagnostic(s)" is an exact-looking claim, and a walk that stopped before whole heaps never emitted what those heaps would have said. Same helper, applied to the nonempty branch; the empty branch already stated coverage.

**`pool_chunk`'s description still called every non-`Allocated` state a free hole** (Codex P2) — correct, and a good catch on staleness: the renderer stopped saying that in b08cf88, but tool discovery would have reintroduced the false use-after-free verdict through the description. Now names `ReusableFree`/`CachedFree` as the freed states and explains `Unreadable` separately, in both the description and the README.

**"Uncovered" framed as proof the address was never pool** (CodeRabbit, Major) — correct, and the same error in the direction that produces false *negatives*. The tool's own output has always said an uncovered address is "either not pool at all, or sits in a region this walk did not reach", and appends the walk report so the reader can tell which; the README and description did not. Both now do.

Two new tests, each mutation-checked — deliberately breaking the implementation makes them fail. (Worth doing here: two tests in the previous round passed under deliberate breakage.) 135 tests pass (117 unit + 18 smoke), fmt and clippy clean.

Also corrected `README.md` "all three read one snapshot" → "all four", a leftover from before `pool_diagnostics` existed.

PR #74, opened 2026-08-07
MessageManager CTF: pool-UAF walkthrough + exploitability PoC (WIP)

glslang:
## What this adds

A walkthrough and PoC for the **MessageManager** CTF kernel driver, exercised end to end on a live
KDNET Windows Server 26100 target with this repo's tools. Its main purpose is to showcase the
kernel-pool tooling (`pool_find_tag` / `pool_chunk` / `pool_census` / `pool_diagnostics`) on a real
bug, and to be honest about how far a modern data-only exploit gets.

- **`docs/messagemanager-walkthrough.md`** — the tour: no-PDB driver reverse (all four IOCTL
  handlers, the `Tgsm` MESSAGE layout, the handle/message lists), the locking use-after-free and its
  Verifier bugcheck, the four Windows-26100 pool-walker breakages that had to be fixed to *see* the
  bug, the `NtQuerySystemInformation` ASLR defeat (validated against the debugger), and a measured
  analysis of the exploit primitives.
- **`examples/messagemanager/mm_exploit.c`** (+ `build.cmd`) — the PoC: `NtQuerySystemInformation`
  infoleaks (kernel/driver base, this `_EPROCESS`/`_KTHREAD`, big-pool), METHOD_BUFFERED IOCTL
  wrappers, and a NonPaged reclaim-measurement mode (`spray pipe|mailslot`) used to probe reclaim
  candidates with the pool tools.

## Status — deliberately honest

This establishes **exploitability, not a shell.** The UAF is confirmed and reproducible, control of
the freed chunk is demonstrated, and the two forward primitives (an unlink write-what-where and an
unconditional arbitrary-free) are pinned to exact instructions. What is **not** solved is the
reclaim: the bug's exploitable fields sit at chunk +0x08/+0x10/+0x18, and every standard medium-IL
NonPagedNx spray measured here (pipe write-data, I/O ring, mailslot) puts a ≥0x10 header above them.
The mitigations (SMEP+SMAP+CET all on) rule out ROP, so the payoff has to be data-only — which is in
hand except for that reclaim. The double-free → type-confusion route that would sidestep it is noted
as future work. Section 6 of the walkthrough lays this out in full.

The exploit code is a work in progress on this branch; the walkthrough is complete.

## Summary by CodeRabbit

* **New Features**
  * Added a Windows example utility with address reporting, driver interaction demonstrations and configurable diagnostic modes.
  * Added a build script to simplify compiling the Windows example with the 64-bit MSVC toolchain.
* **Documentation**
  * Added a comprehensive walkthrough covering debugging, memory inspection, vulnerability analysis and troubleshooting on Windows.
* **Chores**
  * Added ignore rules for generated Windows build artefacts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,9 @@ target
 node_modules/
 package-lock.json
 
+# MSVC C build artifacts (examples/messagemanager exploit)
+*.obj
+*.exe
+
 # Local Claude Code config
 .claude/
--- a/docs/messagemanager-walkthrough.md
+++ b/docs/messagemanager-walkthrough.md
@@ -184,26 +184,47 @@ Each value cross-checks against the debugger: `x nt!PsInitialSystemProcess` prin
 `0xfffff806777c5ab0`, and `lm` gives the same driver base — so the leak plumbing is sound before a
 single byte of the driver is corrupted.
 
-## 6. From UAF to SYSTEM — the plan
+## 6. From UAF to SYSTEM — how far the primitives reach
 
-With a leak and a freed-but-referenced 0x68 chunk, the chain is:
+The intended payoff is **data-only** — the mitigations rule out anything else. CR4 on this box is
+`0xb50ef8`: **SMEP, SMAP, and CET all on**. So a ROP chain to clear CR4 is a dead end (CET faults
+the `ret`s, and there's no instruction-pointer control to begin with — the bug yields *data* writes,
+and kCFG guards indirect calls). The only viable model is to never execute attacker code:
 
-1. **Reclaim** the freed `Tgsm` chunk with a NonPaged spray controlled at `+0x08`/`+0x10` — the
-   `Flink`/`Blink` the Flush unlink writes. (This is the open engineering problem: most medium-IL
-   NonPaged sprays carry a header at the front. The reclaim is *measured* with `pool_find_tag` /
-   `pool_chunk` rather than assumed — the reason the tooling exists.)
-2. **Write-what-where** via the `+0x08` LIST_ENTRY unlink in Flush: `[Blink]=Flink` (with collateral
-   `[Flink+8]=Blink`, so both must be writable — it writes a *pointer*, not an arbitrary value).
-3. **Bootstrap R/W** by setting `KTHREAD.PreviousMode` (`+0x232`) to 0, which makes
+1. **Write-what-where** via the `+0x08` LIST_ENTRY unlink in Flush/SetData: `[Blink]=Flink` with
+   collateral `[Flink+8]=Blink` — it writes a *pointer* to an attacker address, so both operands
+   must be writable kernel addresses.
+2. **Bootstrap R/W** by writing `KTHREAD.PreviousMode` (`+0x232`) to 0, which makes
    `Nt{Read,Write}VirtualMemory` operate on kernel addresses.
-4. **Token steal:** read the System process's token via `PsInitialSystemProcess` and write it into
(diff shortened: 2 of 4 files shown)


glslang:
All three Codex findings addressed in `47c6771` — they were all correct, and two of them would have made the measurement *lie*, which for a tool whose whole job is measuring reclaim offsets is the worst failure mode.

**Reject `datalen < 8`** (P2) — correct. The 8-byte `"MMSPRAY!"` magic is `memcpy`'d unconditionally, so a 1–7 byte payload overflowed the record allocation before any pool was even inspected. `do_spray` now rejects it up front (covers both spray helpers).

**Require a complete `WriteFile`** (P2) — correct, and the more insidious of the two. A short or failed write left nothing resident in pool, yet the helper still returned the server handle and `do_spray` counted the record as queued — so the "N/N records queued" line, and any conclusion drawn from the pool afterward, would be based on records that aren't there. Both helpers now close both handles and return `NULL` unless the full payload was written.

**Print an exact tag** (P2) — correct. `pool_find_tag` takes an exact 1–4 byte tag with no wildcards, so the old `NpF*/Msfs` hint would resolve nothing. Now it prints `NpFr` for the pipe spray, and for mailslot — which (as this branch's own measurement found) keeps no separate reclaimable chunk — it points at the FILE_OBJECT route instead of a tag.

Compiles clean with the host MSVC toolchain. CodeRabbit had no actionable comments; CI was green before these and they're spray-tooling-only.

glslang:
Two follow-up Codex findings addressed in `e9299cd`, both valid robustness gaps in the measurement harness:

- **Reject unknown spray primitives** — `spray mailslots` (a typo) fell through to the pipe spray while the banner printed `mailslots`, which would attribute the observed pool layout to the wrong primitive. Now only `pipe` and `mailslot` are accepted; anything else errors.
- **Guard the handle-array `calloc`** — a large `count` or memory pressure could make it return NULL, after which the first successful record did `held[i] = h` (null write). Now checked before the loop, and `count <= 0` is rejected too.

Compiles clean; CI was green on the prior fix commit. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

glslang:
Addressed in `f1bacbc` — both about the guidance the tool prints after spraying, and both correct:

- **Mailslot hint contradicted the finding** — it claimed the message was reachable via `FILE_OBJECT.FsContext`, but this branch's own measurement (walkthrough §6) found the MSFS FCB data queue empty. Mailslot mode is a **negative control**; the hint now says so.
- **`pool_find_tag` reuses the cached snapshot** — a walk from before the spray omits the new records, so the hint now asks for `refresh: true`. (Good catch on a gotcha from the snapshot caching itself.)

Compiles clean. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

glslang:
Addressed in `bbe0602`: `do_demo` now returns nonzero when `SetData` or `Delete` fails, so a script using that mode to validate the Create/SetData/Delete round trip can't report success on a failed driver op. Compiles clean. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

glslang:
Addressed in `52f7b15`:

- **Bound the pipe payload** — pipe `datalen` is now capped at the pipe buffer (`MM_PIPE_BUFFER` = 0x1000, shared with `CreateNamedPipe` so they can't drift); a larger synchronous write would block with no reader and hang before the banner.
- **Bail on an empty spray** — if every record failed (`ok == 0`), free the array and return failure instead of announcing a hold and sleeping 10 minutes on an empty pool.

Compiles clean. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #100, opened 2026-08-11
Reclaim the freed MessageManager chunk without the debugger

glslang:
## What

The MessageManager walkthrough's RIP proof (§8) depended on a **KD-forced allocation return** to drop a fake MESSAGE into the freed `0x68` `Tgsm` chunk — because `do_rip`'s reclaim source is a **paged** `Tfub` buffer, the wrong pool for the NonPaged `Tgsm`. This PR shows the **reclaim step no longer needs the debugger**.

New `reclaimprobe` mode separates the reclaim question from the race: free a *run* of bare `Tgsm` chunks cleanly, capture each freed address from the driver handle list, spray a NonPaged candidate on CPU 0, and read each slot back.

## The finding

The reclaim's real constraint is **allocation context, not size**: the freed `0x80` slots are reused live by system `Event` objects, which allocate through the *generic* `ExAllocatePool2(NonPagedNx)` path the driver's `Create` uses. Measured live on **26100.32995**:

| Candidate | Result |
|---|---|
| pending pipe write | `NpFr` (`0xb0`, +0x40) — wrong bucket/offset. Miss. |
| `FSCTL_PIPE_TRANSCEIVE` (`0x11C017`) | `METHOD_NEITHER` — no SystemBuffer at all. |
| AFD context (`IOCTL_AFD_SET_CONTEXT`) | real `+0x00` NonPaged spray, but `AfdC` is a **separate context** — 0/16 reclaimed. |
| **pending `FSCTL_PIPE_WAIT`** (tag `IoSB`) | **reclaims** — I/O-manager buffered SystemBuffer on the generic context. **8/16** freed targets reclaimed, `MMWAIT!!` at `+0x00`, **no debugger**. |

Two plumbing points make `PIPE_WAIT` work: issue via **`NtFsControlFile`** (not `DeviceIoControl`, which sends `IRP_MJ_DEVICE_CONTROL` and gets `ERROR_INVALID_FUNCTION` for a NAMED_PIPE-device FSCTL), and open the NPFS root **async + as a directory** so the wait parks `STATUS_PENDING` and keeps the buffer live.

## Boundary (what's left)

The `IoSB` buffer controls `+0x00` (RefCount) and `+0x10..` (Name → **Buffer `+0x18`**, the arbitrary-free primitive) but **not** Flink `+0x08` (that's `NameLength`, a validated length, not a pointer). So:

- **Removed:** the KD reclaim assist — the reclaim is natural now.
- **Newly debugger-free:** the data-only **arbitrary-free** route (`ExFreePoolWithTag(msg+0x18)`).
- **Still KD-only:** the §8 *unlink* RIP's `+0x08` store.

Walkthrough §6/§8 rewritten to the positive result. No privilege-escalation claim; the next phase (weaponizing the arb-free) is deliberately not in this PR.

## Testing

`reclaimprobe [window][spray][hold][targets][wait|afd|pipe]` (default `wait`), driven live over KDNET with the debugger reading each freed address out of band. Harness builds clean with the sibling `build.cmd`.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Message Manager walkthrough with revised grooming measurements, allocation guidance, buffer details, and clarified conclusions.
  * Documented the distinction between debugger-assisted and natural reclaim observations.

* **Tools**
  * Added a `reclaimprobe` mode to compare supported memory-reclaim scenarios.
  * Added validation, CPU-affinity handling, retained-allocation inspection, and clearer usage guidance.
  * Extended networking support and improved reporting for probe results.

### DIFF
--- a/docs/messagemanager-walkthrough.md
+++ b/docs/messagemanager-walkthrough.md
@@ -200,7 +200,7 @@ Windows Server 26100.32995 were:
 | Candidate | Live measurement | Result |
 |---|---|---|
 | NPFS write-data (`NpFr`) | physical blocks were `0xa0`; payload followed an NPFS header | wrong bucket and offset |
-| Pending METHOD_BUFFERED requests (`IoSB`) | 784/1024 exact-`0x68` SystemBuffers stayed parked | target remained `ReusableFree` |
+| Pending pipe writes ("IoSB" attempt) | payload lands in an `NpFr` block (`0xb0`, data at `+0x40`), **not** a distinct `0x68` SystemBuffer | wrong bucket and offset; freed slots stayed `ReusableFree` |
 | Event objects | requested `0x68` and appeared in mixed `0x80` LFH blocks near `Tgsm` | useful density groom, unreliable reclaim |
 | Retained SetData buffers (`Tfub`) | 1024 exact-`0x68` driver copies | target remained free: SetData uses `POOL_FLAG_PAGED` (`0x100`) |
 | SetData with KD-patched tag and flags | `Tgsm` plus `POOL_FLAG_NON_PAGED` (`0x40`) | still missed; call-site/subsegment selection remained different |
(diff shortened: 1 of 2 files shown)


glslang:
The `release_pending_pipewait` finding (free-while-pending -> heap corruption) is fixed in follow-up #101, commit `1c5c8c1`: it now cancels, waits INFINITE for the completion event, and only frees on confirmed terminal completion (leaks the record otherwise). #100's branch is merged, so the fix rides the stacked #101.

PR #114, opened 2026-08-12
feat(triage): read a bug check as fields, not as 150 lines of !analyze

### LINKED ISSUE #104: Add a crash-triage / bugcheck-summary tool for dumps and live bugchecks
## Problem

The debugger-free MessageManager work (PR #102) is proven by reading crash dumps: fire the UAF from user mode, let the guest bugcheck + auto-reboot, then classify the minidump. Doing that classification currently means opening the dump and running `!analyze -v`, which returns ~150 lines (every `Hypervisor.*` key, timing, blackbox flags, …) to extract four fields I actually needed each time:

- bugcheck code + parameters,
- faulting instruction as `module+RVA` (the driver frame, since there is no PDB),
- the offending pool tag / `FREED_POOL_TAG`,
- `PROCESS_NAME`.

Across ~10 crash/reboot cycles I ran `!analyze -v` repeatedly and hand-parsed the same handful of lines to tell an *intended* fire (`MessageManager!SetData` → `ExFreePoolWithTag`, `+0x1654`) from an incidental one (Flush unlink AV `+0x14e9`; a `Tfub` LFH double-free `+0x1668`).

## Proposed shape

A `bugcheck` / `crash_triage` tool (dump sessions, and live-kernel sessions that are stopped at a bugcheck) returning structured fields:

```json
{
  "code": "0x13A", "name": "KERNEL_MODE_HEAP_CORRUPTION",
  "parameters": ["0x11", "0x...", "0x...", "0x0"],
  "faulting_frame": { "module": "MessageManager", "rva": "0x1654", "symbol": null },
  "faulting_pool_tag": "Tfub",
  "process_name": "mm_exploit.exe",
  "top_frames": [ "nt!KeBugCheckEx", "nt!RtlpHpVsContextFree", "nt!ExFreePoolWithTag", "MessageManager+0x1654" ]
}
```

Module-relative framing matters here because the interesting driver has **no PDB**, so `!analyze`'s own `Failure.Exception.IP.Module` is often wrong (it named `mpsdrv` in an earlier run) — a `module+RVA` computed from the loaded base is the reliable signal. This pairs naturally with #84 (structuredContent) but is a distinct, high-value convenience for the driver/crash-dump workflows this server targets.

glslang:
Closes #104.

> [!IMPORTANT]
> **Blocked on [glslang/win-kexp#96](https://github.com/glslang/win-kexp/pull/96)**, which adds the four DbgEng primitives this is built on. CI here will fail until that merges and `Cargo.lock` is bumped with `cargo update -p win-kexp` (a follow-up commit on this branch). Everything below is verified locally against a `[patch]`ed win-kexp.

## The problem

Classifying a crash meant `!analyze -v` and hand-reading ~150 lines of it for four things: the bug check code and its parameters, the driver frame, the pool tag, and the process. Fine once; unreadable on the tenth crash/reboot cycle of the debugger-free exploitation loop in #102.

## `crash_triage`

```jsonc
crash_triage {}                      // a kernel crash dump, or a live kernel that bug checked
crash_triage { "analyze": false }    // engine reads only — fast
crash_triage { "frames": 64 }        // deeper walk
```

```text
BUG CHECK: 0x9f DRIVER_POWER_STATE_FAILURE
  Arg1: 0x0000000000000003  A device object has been blocking an IRP for too long a time
  Arg2: 0xffffe284ffe59060  Physical Device Object of the stack
  Arg3: 0xffffd38c2d84f580  nt!TRIAGE_9F_POWER on Win7 and higher, otherwise the Functional Device Object of the stack
  Arg4: 0xffffe2850787bc20  The blocked IRP
PROCESS: System
FAULTING FRAME: none — every one of the 7 captured frames is in the kernel image or the HAL…
FAILURE BUCKET: 0x9F_3
!analyze blamed: Unknown_Module

STACK (7 frames):
  00 nt!KeBugCheckEx  [nt+0x4f8450]
  01 nt!PopIrpWatchdogBugcheck+0x1f5  [nt+0x5ca7fd]
  …
```

…plus the same answer as `structuredContent` against a declared `outputSchema`.

## Two provenances, deliberately kept apart

**Engine reads** — the code and parameters (`ReadBugCheckData`), the stack, each frame's module, and the crashing process (out of the current `_EPROCESS`, where `!analyze` reads `PROCESS_NAME`).

**`!analyze`'s conclusions** — the pool tag, failure bucket, blamed module and per-parameter explanations exist nowhere else, so they are extracted from its output (the one place in this server where a structured field is scraped from debugger text) and confined to an `analysis` object a consumer can tell apart.

That split is the point rather than tidiness. `!analyze` attributes a crash to a module by heuristic and is often wrong for a driver with no PDB — which is every driver this workflow cares about; it named `mpsdrv` for a crash in `MessageManager`. So frames are attributed **independently**, from the engine's own load bases, and `faulting_frame` is the topmost frame outside `nt`/`hal`, named `module+RVA` — an offset comparable across reboots, which is exactly what tells an intended fire (`+0x1654`) from an incidental one (`+0x14e9`). `!analyze`'s own answer rides along under `analysis.module_name`, and the text prints the disagreement when there is one.

## The answers that are not fields

- **`faulting_frame: null` is a finding, not a gap.** `faulting_frame_note` says whether the whole captured stack really is in the kernel image — a `0x9F` watchdog fires on an idle CPU's timer DPC, so the culprit is not on that stack at all — or whether the walk merely hit its cap, which `frames` raises (`frames_truncated` carries the same fact as a value).
- **Only the bug check itself can fail the call.** An unreadable stack, an unnameable process, or a missing `!analyze` each come back as an absent field *with the reason*: a crash is still a crash without them, and failing the call would throw away everything the dump did give up.
- **A target that did not bug check says so in those terms.** Code 0 (a live kernel that has not crashed yet; a dump that is not a crash dump) and a user-mode session that has no bug check data at all each get a message about the target, not an HRESULT.
- **Bug check names come from a table here**, so `analyze: false` and an engine with no extension DLLs still name the crash; a code the table does not know falls back to `!analyze`'s header line. `!analyze` itself is tried both ways (`!analyze -v`, then the module-qualified `!ext.analyze -v` the bundled engine needs) and the report says which worked.

## Verification

- 14 unit tests in `src/triage.rs` over scripted values: faulting-frame selection, RVA stability across load bases, the all-kernel and truncated-walk notes, the `!analyze` extraction (including the two blocks that *look* like key/value pairs — the indented `KEY_VALUES_STRING` block and the rendered stack), and the text.
- A new debugger-tier smoke test drives it against the checked-in `0x9F` minidump end to end. It asserts `process_name` is `System` specifically because the engine's own `GetCurrentProcessExecutableName` answers `ntkrnlmp.exe` on a kernel target for every process that has ever run — a regression invisible in every other check.
- `tools/list` golden re-recorded (43 → 44 tools); the structured-content contract test covers the new tool on its failure path too.

Docs: README (tool table, structured-results table, limitations), the `crash-dump` skill playbook, `docs/crash-dump-walkthrough.md` (a new §2 with the real output), and `docs/smoke-test.md`.

## Summary by CodeRabbit

* **New Features**
  * Added the `crash_triage` tool for concise crash-dump summaries.
  * Reports bug-check details, stack frames, process information and optional analysis results.
  * Supports configurable frame limits, analysis control and session selection.
  * Provides structured and human-readable output, including faulting-frame attribution and truncation status.
* **Documentation**
  * Added usage guidance, workflow instructions, limitations and smoke-test coverage.
* **Tests**
  * Added coverage for crash metadata, stack attribution, process details, analysis fallback and truncated output.


glslang:
Rebuilt on the merged win-kexp (`Cargo.lock` → `1892acf`) and addressed the review in 2d2343b.

**The three budget findings were one mistake in three places, and all are real.** `run_analyze` started its clock *after* the engine reads, so an `!analyze` could be handed nearly the full patience the triage had already spent — and those reads are genuinely slow when resolving a frame's symbol pulls a PDB over the network. The clock now starts at the first read. Separately, `watchdog_budget_ms` floors at the 15s headroom (deliberately — a command dequeued past its deadline still has to be bounded by *something*), so two attempts taken unconditionally could between them run two floors past the caller; the fallback spelling is now skipped when the patience is already spent, and the report says that is why.

**Truncation was the sharpest finding.** `!analyze` prints its summary block early, so a watchdog-truncated run passes `looks_analysed` and would have been reported as complete — making a `pool_tag` that was never reached indistinguishable from one `!analyze` decided the bug check does not have. There is now an `Analysis::Truncated` state that keeps the fields it did print, sets `analysis.truncated`, and explains what a missing field does and does not mean. Covered by a new unit test.

**`CrashTriage` missing from `an_op_that_carries_a_deadline_hands_it_to_the_pump`** — fixed, and thank you: that test exists precisely to catch a variant carrying `patience_ms` that the pump is never told about, which is the bug `EngineOp::Pool` shipped with. **Case-insensitive module comparison** — fixed, with a test; `is_kernel_image` was already the convention.

**On the nullable `faulting_frame`:** the field is *absent*, not null — every optional field in this server's structured results uses `skip_serializing_if`, and `structured.rs` has a test asserting that ("absent rather than null: a consumer reading `waited_ms` gets 'no answer', not '0ms'"). So the fix was to the prose, in all three places that said "null", and the walkthrough's JSON now shows the omission explicitly. Both places also now say what to *do* about it, which for a `0x9F` watchdog is to go to the bug check arguments — §4 of the walkthrough.

**On the smoke tier asserting more than the docs promise:** agreed, and resolved toward the doc rather than the test. The exact bucket string and a parameter-note count of four are computed by whichever `ext.dll` the host has, so pinning them would fail the tier on a different WinDbg rather than on a change here. It now checks shape — notes stay positional (`<= 4`), the bucket is derived from the code (`starts_with("0x9F")`) — and both only when the run was not truncated.

Verified: `cargo fmt`, `cargo clippy --all-targets`, and `WINDBG_MCP_SMOKE_DUMP=1 cargo test` (254 unit + 32 integration) all clean against the merged dependency.

glslang:
Second round addressed in fc6dafb. All seven findings were valid.

**`!analyze` now runs *before* the engine reads** (codex, "preserve debugger context"). I measured it first: on the sample `0x9F` it moves nothing — that bug check has no exception context to switch to — but the concern is real for the class that does (`0x8E`/`0x7E`/`0x3B` `.cxr` to the context record, and those are the common driver crashes), so one green data point doesn't clear it. Reordering fixes it on the merits rather than as a workaround: reading *after* the analysis means the frames describe the thread `!analyze` blamed, which is the stack actually worth having, and two consecutive triages now re-select the same context before reading instead of the second inheriting whatever the first left. The residual — that `analyze: true` may leave the session's context where `!analyze` put it, exactly as running it through `execute` would — is now documented on the tool and in the README, with `analyze: false` as the way to touch nothing.

**The exhaustion check now guards every attempt** (codex). The `attempt > 0` condition was straightforwardly wrong: the reads above can spend the whole patience on their own, and `watchdog_budget_ms` floors at the headroom, so the *first* `!analyze` could start with nothing left and still run a floor past the caller. Folded in with the outside-diff finding — a skipped-for-time analysis previously blamed a missing `winext\ext.dll`, sending the caller to look for a file sitting exactly where it should be. The two causes are now reported separately.

**Truncation carries its cause** (CodeRabbit). `OnRequest` and `Deadline` both land there and the advice is opposite; "raise the timeout" is useless to someone who called `interrupt` on purpose.

**The framework-layer finding was the sharpest of the round** (codex). `nt -> Wdf01000 -> MyDriver` named the framework, for every KMDF bug in existence. KMDF and Driver Verifier are now passed through, with a deliberately *short* list — `ndis`, `storport`, `fltmgr` are **not** in it, because a crash inside them is as often theirs as their caller's, and skipping them would hide the answer rather than find it.

But the list was the smaller half. The real defect was the renderer telling callers to *prefer* the frame over `!analyze`'s attribution, and you're right that this over-claimed. Only the **offset** is computed — the engine says which image holds the address and where it is loaded. Which frame is *guilty* is a positional heuristic, and `!analyze`'s is a different heuristic; they fail in different ways. The text now reports the disagreement, says neither is authoritative, and points at `STACK`, where every `module+RVA` is sound whichever guess is right. Same correction applied to the module docs, the structured field docs, the README and the skill.

**`PROCESS_NAME` from a truncated analysis** (CodeRabbit) — fixed; demanding a field the tool explicitly says may be missing would have failed the tier for the one behaviour it exists to allow. **README no-bug-check wording** (CodeRabbit) — fixed; "reports code 0" described the engine, not the tool, which refuses with a message naming which of the two cases applies.

Separately, and not from the review: **the live-kernel tier now covers `crash_triage`.** I had said in passing that the ignored tiers don't touch this code path, and that was wrong — this tool was written *for* the debugger-free exploitation loop, which spends most of its time attached to a kernel that has not bug checked yet. A dump can never be in that state: it either is a crash dump or is not, while a live kernel moves between the two. So it is the only place both arms of the refusal are reachable, the only place a kernel target can be shown *not* to take the user-mode arm, and the only place "a triage that came too early costs the session nothing" means anything.

Verified: `cargo fmt`, `cargo clippy --all-targets`, `WINDBG_MCP_SMOKE_DUMP=1 cargo test` (255 unit + 32 integration) clean. The live-kernel assertions need a real KDNET target and have not been run here.

glslang:
Third round addressed, and then the tool was pointed at the CTF it was written for.

## The three findings (all valid)

**Any cut-short now ends the attempts, readable output or not.** An `interrupt` landing on the first `!analyze` before it printed anything recognisable fell through to the ordinary no-analysis arm and started `!ext.analyze -v` — restarting, under a fresh watchdog, the very work the caller had asked to stop. The general form is cleaner than the special case: the fallback spelling exists for *one* failure, an engine on which the command does not resolve, and a command that got far enough to be cut short plainly resolved. That also closes the deadline variant, which the `spent >= patience` guard misses because a deadline fires with the reply's headroom still nominally unspent.

**`frames_truncated` is now established by walking one frame further and discarding it**, exactly as suggested.

**`analyze: false`** no longer claims a context it did not select — documented rather than papered over, since `!analyze` is what re-selects the faulting context.

## Then: five real MessageManager crash dumps

The headline claim holds. `faulting_frame` is `MessageManager+0x1654` on **all five**, at five different load addresses (`0xfffff8020a5e1654`, `0xfffff80561281654`, `0xfffff8070b001654`, …), while `!analyze` says `Unknown_Module` on every one.

It also found **two bugs that every green run had missed**:

- **`process_name` was truncated.** `_EPROCESS::ImageFileName` is 15 bytes, so `mm_exploit_v5.exe` came back as `mm_exploit_v5.`. `!analyze` had the full name on all five, so this PR's own claim that the field was read "where `!analyze` reads it" was wrong. Fixed upstream in [glslang/win-kexp#97](https://github.com/glslang/win-kexp/pull/97) by reading the audit name; the `Cargo.lock` bump follows that merge.
- **`symbol` was reporting a module name as a symbol.** For an address in a PDB-less module the engine answers with the module itself plus a displacement — `module`+`rva` in disguise. Verbatim it rendered `MessageManager+0x1654  [MessageManager+0x1654]`, and set a `symbol` field a consumer would read as "this frame resolved", for the one case the whole tool exists to handle.

Both hid behind the checked-in sample for the same reason: it is a `0x9F` watchdog with no driver frame at all, and its process is `System` — short enough to fit fifteen bytes. **A convenient fixture is not a representative one**, so a driver crash dump is now checked in beside it (`docs/samples/081226-2187-01.dmp`, 687 KB, a `0x13A` out of `nt!ExFreePoolWithTag`). The debugger tier now pins what the other dump structurally cannot: a faulting frame six frames below the top under a stack of allocator internals; its RVA as a literal, because it is a fixed offset into a fixed image; an absent `symbol`; a `pool_tag`; and a process name longer than fifteen characters. Walkthrough §7 shows the same crash with real output.

The **live-kernel tier** passes against the target too — all eight tests, including the refusal a kernel that has not bug checked has to give.

> [!NOTE]
> Still blocked on win-kexp#97 for the `Cargo.lock` bump. The new fixture test asserts the full process name, so CI will be red until that merges.

glslang:
`Cargo.lock` bumped to the merged win-kexp (`74a7595`) in 88a73fe — the debugger tier is green on the full driver-crash fixture, including the process name that needed the audit-name fix.

Fourth review round addressed in 3a3ff37. Both valid, and the second is a cost of my own reordering.

**"Some patience left" was the wrong test.** `watchdog_budget_ms` floors at `WATCHDOG_HEADROOM` — deliberately, since a command dequeued past its deadline still has to be bounded by *something* — so an attempt begun with 3s left is armed with a 15s watchdog anyway, and 12 of those seconds run on a session whose caller has already timed out and can no longer even interrupt it. The floor is what the remaining patience must cover, and that is now the check (`analysis_fits`).

Its test pins the **property** rather than the constant: across the whole range, whenever `analysis_fits` says yes the budget then armed is no longer than the time remaining, and whenever it says no that budget *would* have overrun. So the check neither lets a command outlive its caller nor refuses time it could have used — and both halves stay true if the floor or the arithmetic is ever retuned.

**Moving `!analyze` first made the reads after it unbounded**, which is on me: they are direct engine calls rather than commands, so no watchdog reaches them, and a frame whose symbol has to come from a symbol server can block inside one. While they came first this was harmless (the analysis after them took whatever they left); last, they run after the analysis has spent the budget. Time nothing can bound has to be reserved instead, so the analysis is handed a patience with `TRIAGE_READ_RESERVE` already removed.

That reserve is `WATCHDOG_HEADROOM` again rather than a new number — this file already has an answer to "how long does the part after the bounded work need?", and a second constant beside it is a second thing to keep in step for no reason. With the default 300s timeout it costs nothing; on a much tighter host it is what makes the triage skip its analysis and *say so* rather than run one that outlives the call.

Verified: `cargo fmt`, `cargo clippy --all-targets`, `WINDBG_MCP_SMOKE_DUMP=1 cargo test` (257 unit + 33 integration) clean against the merged dependency.

> [!NOTE]
> CodeRabbit has reported "Review rate limited" on the last three pushes, so its input on those rounds is missing — worth a look before merge if you want its second opinion on them.

glslang:
Fifth round addressed in 870630f. Both valid.

**"Does the watchdog fit" was still the wrong threshold.** `watchdog_budget_ms` subtracts `WATCHDOG_HEADROOM` *precisely* so a command stopped at its deadline has time for the Ctrl+Break to unwind and the answer to travel — and it floors at that same figure. Below twice the headroom the floor wins, and what it overrides is exactly the reserve the reply was relying on: with 15s left the watchdog is armed for 15, so a command that ran to its deadline leaves nothing to answer in. Twice the headroom is the point below which the function stops keeping its own promise, so that is the threshold now.

The test needed only its property restating, which is the value of having written it that way: the armed budget **plus** the headroom must fit, asserted in both directions — never overrunning, and never refusing time it could have used.

**A user-mode frame can never be the faulting kernel driver**, and you're right that the answer was already in hand: `Module::user_mode` came back from the engine and `AttributedFrame` was holding it right up until `frame_info` dropped it. The candidate is now chosen over the attributed frames rather than the rendered ones.

Worth noting the third exclusion is *not* a fourth: an address in **no** module stays a candidate. A freed pool page, an unloaded driver and a corrupted return address all look like that, and all three are what a driver bug leaves behind — the CTF dumps have three such frames each, right below the driver. `is_candidate_culprit` now states all three rules and that distinction in one place.

Regression test covers the full shape: a stack crossing the syscall boundary into `ntdll` and `mm_exploit.exe` names no driver and says "user mode" in the note, the user-mode frames stay in `frames` (excluded from blame, not from the walk), and the same stack with a driver spliced in *below* the boundary still finds it — the rule is about the mode of the module, not about position.

Verified: `cargo fmt`, `cargo clippy --all-targets`, `WINDBG_MCP_SMOKE_DUMP=1 cargo test` (258 unit + 33 integration) clean.

glslang:
Sixth round addressed in e46c862. All five valid.

**`read_only_hint = true` was a promise this tool does not keep**, and you're right that the last round made that worse rather than better: I documented the context mutation in the README and left the annotation claiming otherwise. Nothing here writes to the debuggee, but `!analyze -v` selects the faulting context on the bug checks that carry one and leaves it selected, so two identical `backtrace` calls either side of a triage can differ â€” a change to the environment the other tools read, which is what the hint is about.

`ioctl_trace` already sets the precedent for this exact shape (session state moves, nothing is harmed), so the triage takes the same combination: `read_only_hint = false, destructive_hint = false`. The annotation is per-tool rather than per-argument, so it has to describe the worst case â€” `analyze: false` really is a pure read and cannot say so.

I looked hard at the alternative you offered first, since it is the better outcome: saving and restoring the scope would let the annotation go back. That is `IDebugSymbols3::GetScope`/`SetScope`, a win-kexp primitive that doesn't exist yet â€” so it is noted at the annotation, with the reason rather than just the conclusion. Worth saying what I could **not** verify: the dumps to hand are `0x13A`, `0xC2` and `0x9F`, none of which carries an exception context, and on the `0x9F` I measured `!analyze` moving nothing. So I have no fixture that demonstrates the mutation. That cuts toward the annotation rather than away from it â€” the current one is correct either way, the old one was a claim I couldn't check.

The other four were wording against behaviour, all fixed:

- the playbook promised `analyze: false` "names the driver frame" and that `!ext.analyze` "already switches to the faulting thread" â€” neither unconditional, now stated conditionally;
- README said the audit name is a full image *path*, but the tool returns the leaf, so it is a full image *name*;
- the driver-crash test demanded `PROCESS_NAME` from a possibly-truncated analysis â€” the same guard the other dump's test already had, now applied to both.

Golden re-recorded; the only diff is the one hint. Verified: `cargo fmt`, `cargo clippy --all-targets`, `markdownlint`, `WINDBG_MCP_SMOKE_DUMP=1 cargo test` (258 unit + 33 integration) clean.

glslang:
Seventh round addressed in 5fbbfb0. Both valid.

**An `interrupt` now stops the triage, not just its retry.** You're right that the previous fix was half of one: it suppressed the fallback `!analyze` after an on-request break and then ran the stack walk, every module lookup and the process read regardless — all unbounded, all able to block on symbol loading. So the caller who asked for the session back waited for work that started *after* they asked.

That needed a state rather than an empty vector, because "the walk found nothing" and "the walk was never attempted" are different answers and only the first is about the crash. `Walk::Abandoned` says so in the note and makes `truncated` unrepresentable where it would be meaningless. What had already been read — the bug check, and whatever `!analyze` printed before it was cut — still comes back.

A **deadline**-truncated analysis is deliberately *not* treated this way: its reads are already reserved for (`TRIAGE_READ_RESERVE`), so abandoning them would throw away an answer nobody asked to abandon. Only an on-request break is cancellation.

**A failed module lookup is no longer evidence that a frame is the culprit.** This one is a good catch and it is the same bug as win-kexp#96 one level down: `module_at`'s error was folded into `None`, and `None` is *deliberately* a candidate — a freed pool page, an unloaded driver and a corrupted return address all look like that, and all three are what a driver bug leaves behind. The CTF dumps have three such frames each, directly below the driver frame. So the fold let one unreadable frame pre-empt the genuine driver below it, on the strength of the engine having declined to answer.

They are opposite facts, so they are now opposite states: `Attribution` has three arms instead of an `Option`'s two. win-kexp kept the distinction after #96; this was the place that threw it away on arrival. Test covers both: the same stack with frame 1 as `NoModule` names frame 1, and with frame 1 as `Unknown` names the driver at frame 2 — with its own RVA, not frame 1's address.

Also raised **glslang/win-kexp#98** for the scope save/restore that would let this tool be `read_only_hint = true` again, and pointed the annotation at it. The issue records what I could *not* measure: no dump to hand carries an exception context, so the mutation the restore is for cannot yet be demonstrated — and this pair of repos has already had one confidently-asserted API fact turn out backwards (`GetModuleByOffset`'s not-found HRESULT), so it says measure first.

Verified: `cargo fmt`, `cargo clippy --all-targets`, `WINDBG_MCP_SMOKE_DUMP=1 cargo test` (260 unit + 33 integration) clean.

glslang:
Eighth round addressed in 2db7167. All four valid.

**An interrupt that printed nothing readable is still an interrupt** — and you're right that the last fix made cancellation depend on timing. It derived the guard from `Analysis::Truncated`, which only exists when `!analyze` got far enough to print something parseable; a Ctrl+Break arriving a moment earlier produced `Unavailable`, the guard did not fire, and the unbounded reads ran anyway. So whether the caller's interrupt was honoured turned on exactly when it landed, which is the one thing a caller cannot aim. There is an `Analysis::Interrupted` state now, and both count.

**The engine's own pending flag is checked beside it**, which is the better half of the finding: one interrupt can arrive in two places. The analysis state sees a break that landed *inside* the `!analyze`; `GetInterrupt` sees one that landed anywhere else — after the analysis returned, or during a triage that never ran one (`analyze: false`), neither of which the analysis state can know about.

The three doc findings were all the same shape — the playbook promising more than the tool guarantees:

- **the `!analyze` fields need the bundled `winext\` extensions**, and it did not say so. Now: without them `analysis.ran` is false *with a reason*, everything read from the engine is unaffected, and `analyze: false` asks for that deliberately rather than trying and reporting the failure.
- **the positional-attribution sentence** said a stack "names that layer" without saying what a reader should do about it. Rewritten.
- **`faulting_frame` can be absent for four reasons and it listed two.** Only two are findings about the crash — an interrupted or failed walk is a fact about the *call*. All four are enumerated now, under an instruction to read `faulting_frame_note` before concluding anything, which is the entire point of that field existing.

Verified: `cargo fmt`, `cargo clippy --all-targets`, `markdownlint`, `WINDBG_MCP_SMOKE_DUMP=1 cargo test` (260 unit + 33 integration) clean.

glslang:
Ninth round: one new finding, fixed in 01c55c1. The other four in this round were re-raised against the previous commit — codex cites blob `5fbbfb0`, and all four were addressed in `2db7167` (the `Analysis::Interrupted` state, the `GetInterrupt` check beside it, and the three playbook items). No action needed on those; happy to re-check if anything still looks unaddressed at HEAD.

**The abandoned-walk note had lost its `\` line continuations**, so the source indentation stayed in the string and `faulting_frame_note` carried two fourteen-space runs mid-sentence. A good catch, and an unusually invisible one: it survives every assertion that checks for a substring, and a diff renders it as ordinary wrapped text. My own fault — I wrote that literal through a script that ate the backslashes.

Asserted now in the test that produces the note: a run of spaces in it fails. Cheap, and worth having given it has happened once.

While fixing it I scanned every string literal in the files this PR touches for the same shape. One other hit, **not from this PR**: the "Not interrupted…" message for a batch running its rollback (`src/worker.rs:861`, from `392a52d`) has the same embedded runs. Left alone here since it is unrelated to crash triage and this diff is long enough — flagged rather than folded in, and easy to fix separately.

Verified: `cargo fmt`, `cargo clippy --all-targets`, `WINDBG_MCP_SMOKE_DUMP=1 cargo test` (260 unit + 33 integration) clean.

PR #116, opened 2026-08-13
crash_triage puts the scope back, and is read-only again

glslang:
Downstream half of [win-kexp#98](https://github.com/glslang/win-kexp/issues/98), whose `ScopeGuard` merged in [win-kexp#99](https://github.com/glslang/win-kexp/pull/99). `Cargo.lock` moves the pin to it.

## What changes

`crash_triage` takes a scope guard for the whole call, so the `!analyze -v` it runs by default no longer costs the caller their selected scope, and `read_only_hint` goes back to `true`. Nothing about the target reads differently.

## The reason in the old comments was wrong, and it was wrong here in three places

This repo said — in `worker::crash_triage`, in the tool description, in `structured::CrashTriage::frames`, in the README and in the `crash-dump` skill — that `!analyze -v` selects the faulting context on the bug checks that carry one and leaves it selected. Measured on four targets, it does not:

| target | scope before | scope after `!analyze -v` |
|---|---|---|
| `0x13A` kernel heap corruption (checked-in sample) | `.frame 3` | frame 0 — **reset** |
| `0xD1` driver fault (produced for this) | `.frame 3` (`myfault+0x1e35`) | frame 0 — **reset** |
| `0x9F` power-state watchdog (checked-in sample) | `.frame 3` | frame 0 — **reset** |
| user-mode access violation | `.ecxr` context | default — **reset** |

From a *default* scope it appears to move nothing, which is why this went unnoticed: the default is the analysis's own fixed point.

What it does move is the implicit **thread**, and it puts that back. The `0x9F` is where that is visible, because the thread it blames is not the one the dump opens on:

```
BEFORE      Implicit thread is now ffffe284`fe4e7040   nt!KeBugCheckEx …
!analyze    Implicit thread is now ffffe284`fe4dd040   <- FAULTING_THREAD, mid-analysis
AFTER       Implicit thread is now ffffe284`fe4e7040   nt!KeBugCheckEx …   (identical stack)
```

So the ordering (`!analyze` first, reads after) is still load-bearing, for a different reason than the comment gave: the analysis *normalises* a session the caller had navigated back to the default before the stack is walked, which is what makes two triages of one session agree. The reported stack is the default context — the crash, on a crash dump — never "the thread the analysis blamed". All five places now say that.

## Test

The smoke tier checks the promise, not the annotation: `.frame 3` → `crash_triage` → the same frame line afterwards, compared as text so it does not depend on symbols. **Frame 3 deliberately**: a check that started at frame 0 would pass whether the scope was restored or merely reset — which is exactly the trap that hid this behaviour for four months.

Mutation-checked: dropping the guard fails it, with `.frame` back at `nt!KeBugCheckEx`.

`UPDATE_GOLDEN=1` re-recorded `tests/golden/tools_list.json`; the diff is one line, `crash_triage`'s `readOnly` flipping to `true`.

Full `WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke`: 33 passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Crash triage now preserves the selected debugger context after analysis, including when interrupted.
  - Subsequent register and backtrace views remain consistent with the caller’s original selection.
  - Analysis uses the default crash context without unexpectedly changing the active session state.

- **Documentation**
  - Clarified context handling, read-only behaviour, and frame selection in the crash triage documentation.

- **Tests**
  - Added coverage confirming selected frame restoration after crash analysis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and a user-mode access violation). The implicit *thread* it does move — visibly, on the `0x9F`,
   where the thread it blames is not the one the dump opens on — and does put back.
 
-  So the stack `crash_triage` reports with `analyze: true` is the target's default context (the
-  crash, on a crash dump) rather than "the thread the analysis blamed", and it is that regardless
-  of where the caller had navigated, which is what makes two triages of one session agree. The
+  So the stack `crash_triage` reports is the target's default context (the crash, on a crash dump)
+  rather than "the thread the analysis blamed", regardless of where the caller had navigated —
+  which is what makes two triages of one session agree. That normalisation is the analysis's own
+  side effect, so it holds exactly when the analysis ran: with `analyze: false`, or when it is
+  skipped for want of time or an `ext.dll`, the walk describes the selected context instead, and
+  `analysis.ran` is what tells them apart. The
   smoke tier checks the promise from a scope the analysis would otherwise discard — frame 3, since
   a check starting at the default would pass whether the scope was restored or merely reset.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,9 +27,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   So the stack `crash_triage` reports is the target's default context (the crash, on a crash dump)
   rather than "the thread the analysis blamed", regardless of where the caller had navigated —
   which is what makes two triages of one session agree. That normalisation is the analysis's own
-  side effect, so it holds exactly when the analysis ran: with `analyze: false`, or when it is
-  skipped for want of time or an `ext.dll`, the walk describes the selected context instead, and
-  `analysis.ran` is what tells them apart. The
+  side effect, so it holds exactly when the analysis *completed*: with `analyze: false`, when it is
+  skipped for want of time or an `ext.dll`, or when the deadline cut it short before the reset it
+  does partway through its output, the walk describes the selected context instead — `analysis.ran`
+  and `analysis.truncated` are what tell those apart. The
   smoke tier checks the promise from a scope the analysis would otherwise discard — frame 3, since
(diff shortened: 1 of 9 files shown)


glslang:
Taken — good catch, and it was a real hole in the *documentation* this PR added rather than in the behaviour it changed.

`run_analyze` declines to start when the patience is already spent, and both spellings can fail to resolve on an engine with no `ext.dll`. In either case nothing resets the scope, so the walk describes the caller's selected context — while the wording I added promised the default context for `analyze: true` flatly. Now qualified against `analysis.ran` in all five places (tool description, `worker::crash_triage`, `structured::CrashTriage::frames`, README, `crash-dump` skill).

I did not take the other half of the suggestion — normalising explicitly before the walk. That would need a "reset to default scope" primitive in win-kexp and, more to the point, it is a behaviour change: the normalisation has always been the analysis's own side effect rather than something this tool promised, and this PR is about restoring what a triage disturbs, not about adding a new guarantee. Worth its own issue if the unconditional version is wanted.

glslang:
Taken — and it lands on the fix I made for the previous finding, which was too loose.

`Analysis::Truncated` reports `ran: true` on purpose (what it printed is real, and `truncated` carries the qualification), and a run the **deadline** cut short is deliberately not a cancellation — `Analysis::interrupted()` is false for it — so the reads still happen. On top of that the reset is something `!analyze` does *partway through its own output*: on a user-mode AV it prints `CONTEXT: (.ecxr)` and then `Resetting default scope` well before the end. So a truncated run may or may not have reached it, and nothing on this side can tell which.

The documented condition is now `ran` **and not** `truncated`, in all five places.

Still declining the "explicitly normalize before the walk" half, for the same reason as before plus a sharper one: the cases it would cover are exactly the ones where `!analyze` does not run, so on any host with a working `ext.dll` the new behaviour has nothing to demonstrate — it would be a claim the test suite cannot reach. If you want `crash_triage` to normalise the context itself rather than inherit the analysis's side effect, that is a real feature (`crash_triage` would then answer the same thing with `analyze: false` too) and deserves its own issue rather than riding in a PR about putting the scope back.

PR #117, opened 2026-08-13
A profiles.json with a UTF-8 BOM is a valid one

glslang:
Found while configuring a real KDNET profile on this host, following the advice `attach_kernel` itself prints.

`%USERPROFILE%\.windbg-mcp\profiles.json` is the one config file this server asks a Windows user to write by hand. Windows PowerShell 5.1's `Set-Content -Encoding utf8` — the obvious way to write it — emits a **UTF-8 BOM**, and `serde_json` rejected the file whole:

```
The configuration was read with problems, which may be why the profile you want is missing:
- C:\Users\Admin\.windbg-mcp\profiles.json is not valid JSON (expected value at line 1 column 1)
```

The JSON was perfect. The message reads as "you wrote it wrong" to someone who had just done exactly what the tool told them to, and the fix on their side is to know that `Set-Content -Encoding utf8` is not UTF-8 in the sense anyone means.

So the BOM is skipped before parsing. Three bytes, and the config becomes writable with the platform's default text writer.

The test writes the file the way PowerShell writes it, rather than asserting on a hand-built string, and is mutation-checked: with the skip removed it fails with the same message the user saw.

Verified end to end afterwards — `attach_kernel {}` lists `ctf-vm`, and `attach_kernel { "profile": "ctf-vm" }` attaches to the live KDNET target with the key never entering the request (the session reports `profile "ctf-vm" (net:port=50000,key=<redacted>)`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Profile JSON files with a leading UTF-8 byte order mark (BOM) are now accepted instead of being rejected as malformed.
* **Documentation**
  * Added a changelog entry documenting support for BOM-prefixed profile files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- A `profiles.json` with a **UTF-8 BOM** is read rather than refused. Windows PowerShell
+  5.1's `Set-Content -Encoding utf8` — the obvious way to write the one config file this
+  server asks a Windows user to write by hand — puts a BOM in front, and `serde_json`
+  rejected the whole file with `expected value at line 1 column 1`: a message that reads as
+  "your JSON is malformed" about a file whose JSON is perfect. Found by configuring a real
+  KDNET profile and having the first attempt refused.
+
 ## [0.8.0] - 2026-08-13
 
 ### Changed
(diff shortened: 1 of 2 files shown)


glslang:
Taken, from both — the block was duplicated. Mutation-checking the fix removed the `strip_prefix` line but not the comment above it, and restoring it put the whole thing back. Running the tests after the revert is what missed it: the code was correct, so they passed and I never re-read the hunk. One copy now.

PR #128, opened 2026-08-15
MessageManager: from a pool UAF to SYSTEM, and the debugger-free boundary

glslang:
Advances the MessageManager CTF exploitation example (`examples/messagemanager/mm_exploit.c` + the walkthrough), driving the pool UAF from a debugger-assisted RIP proof all the way to a composed SYSTEM escalation, and localizing exactly what a *fully* debugger-free fire still owes the debugger. No changes to the MCP server (`src/`); this is docs + the exploit example + its recordings.

## What lands

**Reclaim, debugger-free and medium-IL.** `FSCTL_PIPE_LISTEN` is the reclaim primitive whose `+0x08` is attacker data rather than a validated length — swept out of 18 candidates, verified verbatim in the reclaimed `0x80`/`IoSB` slots, measured reclaiming freed `Tgsm` at 15/31. Wired into `reclaimprobe` and `driftarb`.

**The unlink write-what-where lands.** Flush's `RemoveEntryList` is a pointer install; a deterministic KD-placed test shows the store completing into an attacker-chosen address (not just faulting), and pins the "fake must be first on the list" constraint.

**An arbitrary read, built from the write.** The `_ETHREAD.ThreadName` (`+0x6a0`) gadget: point it at a fake `UNICODE_STRING` and `NtQueryInformationThread(ThreadNameInformation)` returns any kernel address. Read the DeviceObject with it.

**Composed to SYSTEM.** `sysswap` reads the System token through the gadget and installs it into `EPROCESS.Token` — `WIN\Admin` → `NT AUTHORITY\SYSTEM`, KD confirming the swapped token. KD stands in only for placing the two unlink-shaped pointer writes; the read is a real syscall.

**The ordering, cracked.** The debugger-free race was thought to always crash on a dirty node. Measured: that was a high-count artifact — at **count=1 the reclaim reliably wins** (the single stale node is our fake). Setter- and Flush-throttle were both measured and falsified as levers; the number of targets is the right one.

**The fully debugger-free RIP fire, built (`ripfire`) and honestly bounded.** Every mechanism is verified — forged `_DRIVER_OBJECT`, the gadget read, the reclaim win, the self-referential-forged-object trick that persists the install. Across four runs it doesn't land RIP for the reason §8 already documents: the race self-corrupts (`0x13A` at `SetData+0x1654`) faster than the controlled install→trigger completes. That's the precise step KD's sequencing buys in §9, now confirmed at the RIP level.

## Honest boundaries (stated in the walkthrough)

- The kernel-address **leaks are admin-only** on 26100/2025 (a 24H2 hardening — a real `lowuser` gets zeroed pointers), so this chain is **admin → SYSTEM** unless paired with a separate medium-IL info leak. The reclaim is medium-IL; the disclosure is not.
- The fully debugger-free **RIP fire** remains blocked by the §8 race-self-corruption.

## Recordings

New `system-proof.cast`/`.gif` (read → SYSTEM), and the `rip-proof.cast` was refit to its own output (it had declared a 32-row viewport for 55 rows) and rendered as a GIF the walkthrough now embeds.


glslang:
Addressed all three Codex findings in `bb563fe`, each verified against the live VM (26100.32995):

- **P1 — baseline big-pool snapshot** (`mm_exploit.c`): `leak_two_scratch` now fails if the baseline `SystemBigPoolInformation` query fails or overflows its table, rather than diffing against an empty baseline and classifying a pre-existing `0x3000` allocation as freshly sprayed. Table grown 512→4096 with an explicit overflow guard. Verified: `sysswap` parks its fake and reaches SYSTEM.
- **P2 — forged MajorFunction offset** (walkthrough): corrected `+0xE0`/`0xE8` → `+0x70`/`0x78`. The trigger is a `CreateFile` (`IRP_MJ_CREATE`, index 0), so `MajorFunction[CREATE]` is at `DRIVER_OBJECT+0x70` — which is what the harness already writes, and what this walkthrough's own `IopfCallDriver` disassembly shows.
- **P1 — ThreadName restore** (`sysswap`/`readgadget`/`ripfire`): after the gadget read, `_ETHREAD.ThreadName` points at the big-pool `IoSB` fake; teardown would free it → `0x19 BAD_POOL_HEADER`. `NtSetInformationThread` can't fix it (it frees the same fake), so the harness now prints a required KD restore step (`eq <ThreadName> 0`) and holds for it. Verified end to end: after nulling `ThreadName` in KD, killing the SYSTEM proof process left the guest **alive with no teardown bugcheck**.

Issue #81, opened 2026-08-09
Support secret-backed KDNET connection profiles


glslang:
## Problem

`attach_kernel` currently requires a raw DbgEng connection string:

```json
{ "connection": "net:port=50000,key=..." }
```

That puts the KDNET key in the MCP request before the server has any opportunity to redact it. During the MessageManager CTF session, the one supplied key was replicated into 524 records in the client JSONL through messages, tool calls, context snapshots, and compaction summaries.

The client is responsible for its own history, but the server API currently gives it no way to avoid handling the secret.

## Proposed shape

Allow `attach_kernel` to take exactly one of:

- `connection`: the current explicit value, for backward compatibility;
- `profile`: a non-secret name resolved by the server from local configuration or environment.

For example:

```json
{ "profile": "ctf-vm" }
```

Profiles should be resolved only inside the windbg-mcp process. Configuration should remain machine-local and outside the repository.

Also centralize connection-string sanitization so `key=` values cannot appear in server logs, tool errors, session diagnostics, or open reports.

## Acceptance criteria

- A kernel session can be opened using a named profile without placing the key in MCP arguments.
- Supplying both/neither required connection selectors produces a clear validation error.
- Errors and logs redact `key=` values from both explicit and profile-resolved strings.
- Unit/smoke tests use fake values and verify that no secret-shaped value is returned.
- Setup documentation explains local profile configuration and keeps raw connection strings supported.

Issue #77, opened 2026-08-08
pool walk reports 71 diagnostics when there were ~7,700 — the count is post-collapse


glslang:
Measured on the live-kernel smoke tier against Server 26100 (glslang/windbg-mcp#76):

```text
71 diagnostic(s) in 24 categories:
        8x  cannot read LFH subsegment <addr> sparse virtual range at <addr> (valid <addr>
        ...
        1x  ... and 120 more like `region # is only committed through # unreadable space extends # bytes`
        1x  ... and <addr> more like `cannot read LFH subsegment # sparse virtual range at # (valid #`
```

Three things wrong, all from the same cause.

**The total is off by ~100x.** The real count is in the tail — `and 5621 more like rejecting
implausible LFH metadata`, `and 1770 more like cannot read LFH subsegment` — so this walk produced
on the order of 7,700 diagnostics, not 71. win-kexp collapses repeats before handing the list over
(glslang/win-kexp#88), so `summarize_diagnostics` counts *surviving lines*. "71 diagnostic(s)" reads
as a measurement of the target and is an artifact of processing, which is the one thing this
rendering exists to avoid.

**The collapse summaries are re-collapsed as if they were findings.** `1x ... and 120 more like ...`
is a category of its own in the table, competing with real ones for the 25 slots.

**Their counts are masked.** `diagnostic_category` replaces numbers with `<addr>`, so
`and 5621 more` renders as `and <addr> more` — the number that made the line worth printing is the
one destroyed.

The honest fix is upstream: `PoolSnapshotReport` should carry the **true** pre-collapse count
alongside the collapsed list, so this can say "7,731 diagnostics (71 lines after collapsing, in 24
categories)". Everything else here follows from not knowing that number. A local patch that parses
`... and N more` back out of the text would work and should not be done — the count belongs in the
type, not in a string this side re-parses.

Filed against this repo because the misleading line is rendered here, but it needs the win-kexp
field first.

**Acceptance:** the reported total matches what the walk actually produced; collapse summaries are
not counted as categories; and no count is masked into `<addr>`.

PR #129, opened 2026-08-16
Record the session: opt-in JSONL transcripts, and a cast renderer

### LINKED ISSUE #87: Add sanitized JSONL and asciicast session transcripts
*Revised 2026-08-16: the original text was written against three features that have since
shipped (#81, #82, #84). Their requirements are folded in below rather than restated as
proposals, and the event list picks up the execution-control work that landed with them.*

## Problem

A debugger server is in the best position to record session-scoped events accurately, and it
still has no opt-in transcript sink. The only output it produces is the `tracing` stream on
stderr (`src/main.rs`, `RUST_LOG`), which is a log of the server, not a record of the session.

The artifacts available instead have opposite problems:

- the generic client JSONL is roughly 34 MB for 43 user messages, includes prompts and secrets,
  and buries debugger operations inside shell-command source;
- a focused regression transcript contains useful final output but not a structured stream of
  individual MCP requests, stop reasons, mutations, or rollback actions;
- **the asciicasts have to be reconstructed after the fact, and still are.** This is the half
  with fresh evidence: `examples/messagemanager/system-proof.cast` was reconstructed on
  2026-08-15, six days after this issue was filed, and the walkthrough says so in as many words —
  *"its timing is illustrative because live recording was not enabled for the original run"*
  (`docs/messagemanager-walkthrough.md`). Both MessageManager casts carry a *"Reconstructed
  terminal session"* caption.

## Proposed shape

An opt-in server-side JSONL recorder configured by local path/environment, with one event per:

- tool request/result;
- session open/state transition/release;
- execution resume/stop reason;
- breakpoint or memory mutation;
- transaction assertion and rollback;
- timeout, cancellation, and worker failure — specifically including `interrupt` stopping a call
  or a batch, and a worker **process** dying, both of which are session-visible events since
  process-per-session (#61/#62) and `interrupt` (#99) landed.

Include monotonic and wall-clock timestamps, session/request identifiers, elapsed time, and typed
verdicts. Add an optional asciicast-v2 renderer for the human-readable stream.

## What has shipped since this was filed

Three of the original premises are no longer speculative, which narrows the work:

- **#81 (closed 2026-08-09) — KDNET connection profiles.** The redaction this issue asks for is
  already built and tested in `src/kdconn.rs`: `redact()`, `Connection::redacted()`, the
  `<redacted>` mask, whole-parameter-name matching, case-insensitive, and it covers `password=`
  as well as `key=`. Serializing a connection is now a call, not a design problem. Profiles also
  narrow the exposure at the source — with `attach_kernel { "profile": … }` the key never enters
  the request at all — but redaction still matters, because raw `connection` stays supported.
- **#82 (closed 2026-08-10) — transactional batches.** `debug_batch` exists, with assertions, a
  rollback block that runs on every path, and typed verdicts (`COMMITTED`, `BATCH: INTERRUPTED`
  with rollback state). The acceptance criterion below now has a real feature to record rather
  than a hypothetical one.
- **#84 (closed 2026-08-12) — `structuredContent` for typed tools.** Tool results already carry
  structured JSON beside the text, so the sink serializes records that exist rather than
  inventing a second representation of each result.

## Security requirements

- Route connection strings through `kdconn`'s existing redaction before serialization; never
  serialize profile-resolved secrets.
- Make recording opt-in and document that arbitrary debugger output may itself contain sensitive
  target data.
- Write atomically enough that a crash leaves a parseable prefix.

## Acceptance criteria

- A short dump session produces valid JSONL that can reconstruct request order and stop state.
- A live attach transcript contains no supplied KD key, by either path (`connection` or
  `profile`).
- A failed transactional batch records the original failure and the rollback outcome; an
  interrupted one records the interrupt and the rollback state it reported.
- An asciicast generated from the same event stream validates as asciicast v2 — and carries real
  timing, so a walkthrough no longer has to caption its recording as reconstructed.
- Transcript output never shares MCP stdout and cannot corrupt JSON-RPC transport.
- Documentation includes retention and redaction guidance.

glslang:
Closes #87.

## The problem

This server is the only party that sees the whole of a session — the tool call, the session it was routed to, the target's answer, the state the session moved to, the worker that died — and it recorded none of it. What existed instead were artifacts with opposite problems: a client's own log, tens of megabytes of prompts with the debugger operations buried inside shell-command source, and the curated proof records under `examples/`, which are readable and are written afterwards.

One of them says so on its face — *"its timing is illustrative because live recording was not enabled for the original run"* — and a second reconstructed cast was added six days after the issue was filed.

## What this adds

`WINDBG_MCP_TRANSCRIPT=<path>` records one JSON object per line. Unset, which is the default, nothing is written and nothing is spent.

```jsonc
{"v":1,"seq":2,"at":"2026-08-16T07:13:01.643Z","mono_ms":135,"event":"session_open","session":"sess-…","kind":"crash dump","target":"C:\dumps\a.dmp","engine_pid":1656}
{"v":1,"seq":10,"at":"2026-08-16T07:13:04.272Z","mono_ms":2765,"event":"assertion","request":3,"session":"sess-…","step":"steps[2] lm m nt","detail":{"text":"the output does not contain \"nope\""}}
{"v":1,"seq":11,"at":"2026-08-16T07:13:04.272Z","mono_ms":2765,"event":"batch","request":3,"session":"sess-…","outcome":"failed","at_step":2,"committed":false,"rollback_complete":true,"after":"stopped","elapsed_ms":603}
```

The events are the tool call and its result; a session opening, changing state, being released; a wait abandoned, an `interrupt`, a worker process dying; and — derived from each result's **typed** half — where execution stopped, what a `run_to_address` concluded, every breakpoint or memory mutation, each assertion that did not hold, and how a `debug_batch` ended with whether its rollback completed.

`windbg-mcp --render-cast <transcript.jsonl>` renders one as asciicast v2 — each call as a prompt line, its output beneath, the derived facts marked between them. Derived offline from the same JSONL, so a cast can be made from a transcript recorded weeks ago and its timings are the recorded ones, measured on a monotonic clock.

## Four properties it is built around

- **Written by the supervisor, from values.** One writer, so no locking and no interleaving between processes — and a worker's facts still arrive as facts, because they cross the pipe as the typed half of its reply and are read back through `structured` types. Nothing here is scraped out of a rendering, which is the rule #77 was fixed by. A tool this server cannot type — an arbitrary `execute` — contributes its call and its result and no derived event, because guessing what the command did would put a fact in the record that nothing measured.
- **Redacted by the same rule as everywhere else.** `kdconn` grew a scan for arbitrary text beside its connection parser, sharing the one list of secret parameter names and the one mask, so a raw `connection` is recorded as `key=<redacted>` and an argument member *named* like a secret is masked whole, before any tool has one. Scrubbing happens **before** the field cap, not after: cutting first can leave the front of a key behind. Profiles (#81) still keep the key out of the request in the first place; this is the backstop.
- **Bounded, and it says so.** Fields are capped (`WINDBG_MCP_TRANSCRIPT_MAX_FIELD`, 16 KiB default) and a record reports how much it dropped. A transcript that quietly truncated would read as complete, which is worse than not having one.
- **Never the transport.** Standard output stays JSON-RPC; the transcript is a file this module opens, and a test reads the source to keep it that way.

## Enabling change: `debug_batch` answers with values

It was the last tool whose whole answer was a rendering, and the one with the most at stake in being readable by a program — a batch runs precisely when something *mutates* the target, and "did this commit" and "did the rollback put everything back" were answerable only by matching on `BATCH: FAILED` and `rollback: INCOMPLETE` in prose. A transcript recording that prose records the wording, not the verdict.

The worker now builds `structured::BatchReportInfo` from the executor's own `BatchReport` and returns it on both paths; `server::batch_result` reads the verdict back to decide `isError`, and **fails closed** — a caller wrongly told "error" re-reads a report that is right there in the text, while one wrongly told "success" walks away from a target that may still be patched.

> [!IMPORTANT]
> **Visible change for existing clients.** `debug_batch` now declares an `outputSchema` and returns `structuredContent`. The text and `isError` are unchanged. Note the pairing only this tool has: a batch that **ran** answers `status: "ok"` (the report is the answer) on a result flagged `isError`, while `status: "error"` means it never ran. Reading only `isError` cannot tell those apart, and it is the difference between "resubmit" and "check what the target is left holding".

The golden `tools_list.json` is re-recorded; the diff is exactly that one schema.

## Testing

`cargo test` (355 unit + 42 smoke), `cargo clippy --all-targets`, `cargo fmt --check`, and the debugger tier (`WINDBG_MCP_SMOKE_DUMP=1`) all pass. Also driven by hand against the checked-in kernel dump, end to end through the shipped binary.

- **Unit** — the record shapes, the redaction, the caps and their truncation markers, the derived events, the RFC 3339 round trip, and the cast's validity, clipping and tolerance of a torn tail. One test reads this module's own source to keep it off standard output; one pins that a transition to the state a session is already in still restamps `in_state_for` but does *not* record a line (verified against a deliberately broken version, which it catches).
- **Protocol tier** — drives the shipped binary with recording on, reads the JSONL back, checks the calls are in order with a `start` first and a `shutdown` last, that the supplied key is absent from the file's **bytes**, that every stdout line still re-parses as JSON-RPC, and that `--render-cast` produces a valid asciicast v2. A second test pins the `profile` path; a third that a server nobody asked to record writes no file at all.
- **Debugger tier** — the batch verdict is now read from the typed half as well as the prose, and the interrupted-batch test checks the transcript records the `interrupt` as its own cause followed by a verdict whose `outcome` is `interrupted` and whose `rollback_complete` is true.

## What is not covered

No KDNET target on this host, so the "a live attach transcript contains no supplied KD key" criterion is checked against attaches that are **refused** (for their shape, so nothing dials), plus the profile path. A landing attach is safe by construction — the raw string travels in `EngineOp::AttachKernel`, which the recorder never sees, and the session label is masked in `kdconn::select` before a worker is spawned — but that is an argument, not a measurement. Worth one pass on the live-kernel tier before release.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in JSONL recording of debugging sessions, with secret redaction, field-size limits and shutdown tracking.
  * Added offline conversion of session transcripts into asciicast recordings.
  * Added structured `debug_batch` results covering outcomes, rollback status, session state and step details.
* **Documentation**
  * Documented transcript recording, rendering, retention, redaction and batch-result semantics.
* **Tests**
  * Expanded coverage for transcript integrity, sensitive-data handling, rendering and structured batch outcomes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,66 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Opt-in session transcripts, and an asciicast renderer for them**
+  ([#87](https://github.com/glslang/windbg-mcp/issues/87)). Point `WINDBG_MCP_TRANSCRIPT` at a path
+  and the supervisor records what it was asked and what it did, one JSON object per line: the tool
+  call and its result; a session opening, changing state and being released; a wait abandoned, an
+  `interrupt`, a worker process dying; and — derived from each result's *typed* half — where
+  execution stopped, what a `run_to_address` concluded, every breakpoint or memory mutation, each
+  assertion that did not hold, and how a `debug_batch` ended with whether its rollback completed.
+  Unset, which is the default, nothing is written.
+
+  What existed before had opposite problems. A client's own log is tens of megabytes of prompts
+  with the debugger operations buried inside shell-command source; the curated proof records under
+  `examples/` are readable and are written afterwards from memory — one of them says so on its
+  face, captioning its own timing as *"illustrative because live recording was not enabled for the
+  original run"*. The server is the only party that sees the whole of a session, and it was
+  recording none of it.
+
+  - **Written by the supervisor, from values.** One writer, so no locking and no interleaving
+    between processes — and a worker's facts still arrive as facts, because they cross the pipe as
+    the typed half of its reply and are read back through [`structured`](./src/structured.rs)
+    types. Nothing in a transcript is scraped out of a rendering, which is the rule
+    [#77](https://github.com/glslang/windbg-mcp/issues/77) was fixed by. A tool this server cannot
+    type — an arbitrary `execute` — contributes its call and its result and no derived event, which
+    is the honest answer rather than a guess about what the command did.
+  - **Redacted, and by the same rule as everywhere else.** `kdconn` grew a scan for arbitrary text
+    beside its connection parser, sharing the one list of secret parameter names, so a raw
+    `connection` passed to `attach_kernel` is recorded as `key=<redacted>`; an argument member
+    *named* like a secret is masked whole, before any tool has one. Profiles still keep the key out
+    of the request in the first place — this is the backstop for the caller who passed a raw string
+    anyway.
+  - **Bounded, and it says so.** Fields are capped (`WINDBG_MCP_TRANSCRIPT_MAX_FIELD`, 16 KiB by
+    default) and a record reports how much it dropped. A transcript that quietly truncated would
+    read as complete, which is worse than not having one.
+  - **Never the transport.** Standard output stays JSON-RPC: the transcript is a file this module
+    opens, and a test reads the source to keep it that way.
+  - **`windbg-mcp --render-cast <transcript.jsonl>`** renders one as an [asciicast v2] recording —
+    each call as a prompt line, its output beneath, and the derived facts marked between them. It
+    is derived offline from the same JSONL, so a cast can be made from a transcript recorded weeks
+    ago and its timings are the recorded ones, measured on a monotonic clock. `--idle-limit`,
+    `--max-lines`, `--title`, `--width`/`--height` shape it.
+
+  Retention is the operator's: nothing but secrets is masked, and debugger output is the contents of
+  somebody's memory. The README says what to do with one.
+
+  [asciicast v2]: https://docs.asciinema.org/manual/asciicast/v2/
+
+- **`debug_batch` answers with its report as values**, not only as prose — the last tool whose
+  whole answer was a rendering, and the one with the most at stake in being readable by a program.
+  `outcome`, the position it stopped `at`, `committed`, `rollback_complete`, what the session holds
+  `after`, and every step of both blocks with what it changed and whether an assertion was `unmet`.
+  It is built in the worker from the executor's own `BatchReport`, which is what lets a transcript
+  record a transaction's verdict as a fact.
+
+  **Visible change for existing clients**: `debug_batch` now declares an `outputSchema` and returns
+  `structuredContent`. The text is unchanged and so is `isError` — a batch that did not commit is
+  still a tool error carrying the whole report. Note the pairing this tool alone has: a batch that
+  *ran* answers `status: "ok"` (the report is the answer) on a result flagged `isError`, while
+  `status: "error"` means the batch never ran. Reading only `isError` cannot tell those apart.
+
 ## [0.9.0] - 2026-08-14
 
 ### Changed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -206,6 +206,25 @@ Issue `.sympath` alone, or use the **`set_symbol_path`** tool (goes through the 
 `AppendSymbolPath`/`SetSymbolPath` API, immune to the quirk; appends + reloads). When a driver's
 `module!Symbol` names don't resolve, **ask the user for the PDB folder** and apply it with that tool.
 
+## Recording a session while debugging this server
+
+`WINDBG_MCP_TRANSCRIPT=<path>` makes the supervisor write a JSONL record of every tool call, every
+session transition, every timeout and every worker death (`src/record.rs`; the README has the
+format). It is off unless the variable is set, and it is often the fastest way to answer "what
+actually happened in that session" — the `tracing` stream on stderr is prose about the *server* and
+interleaves both roles, while this is values about the *session*, keyed by session and request.
+
+Two things worth knowing when using it here:
+
+- **It records the supervisor's view.** A worker inherits the variable and ignores it (the role
+  check in `main` runs first), so there is exactly one writer and no interleaving. A fact that
+  exists only inside a worker reaches the transcript only if it crosses the pipe as a value — which
+  is the same rule as everything else in `src/structured.rs`, and the reason `debug_batch` grew a
+  typed report.
+- **`windbg-mcp --render-cast <transcript.jsonl>`** turns one into an asciicast. That is the
+  supported way to produce the recordings under `examples/` and `docs/` — the older ones are
+  hand-reconstructed and say so, and a new walkthrough should not add another.
+
 ## Plugin vs. dev build
 
 This project is also installed as a user-scope Claude Code plugin (`windbg-mcp@windbg-mcp`), which is
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -151,16 +151,23 @@ alone will happily validate the wrong guest, which is what the `hostip` comparis
 
 ## Live kernel + driver IOCTL gotchas (learned driving HEVD over KDNET)
 
-**Attach by `profile`, not by connection string.** `attach_kernel { "profile": "<name>" }` resolves
-the connection inside the server (`src/kdconn.rs`), so the target's debug key never lands in a tool
-argument — and therefore never in the session transcript, where one key previously ended up
-replicated across hundreds of records. `attach_kernel {}` lists the profiles this host has.
-Configure one with `WINDBG_MCP_PROFILE_<NAME>` or `%USERPROFILE%\.windbg-mcp\profiles.json`; raw
-`connection` still works for a target nothing is configured for. Note the live smoke tier below is
-one of those: `WINDBG_MCP_SMOKE_KERNEL` is a raw connection string, passed straight to
-`attach_kernel { "connection": … }`, and is deliberately *not* a profile — the tier has to exercise
-the explicit path, and it is a variable in a developer's own shell rather than something a client
-ever sees.
+**Attach by `profile`, not by connection string — always, for any live target.** `attach_kernel
+{ "profile": "<name>" }` resolves the connection inside the server (`src/kdconn.rs`), so the
+target's debug key never lands in a tool argument — and therefore never in the *client's*
+transcript, where one key previously ended up replicated across hundreds of records.
+`attach_kernel {}` lists the profiles this host has. Configure one with
+`WINDBG_MCP_PROFILE_<NAME>` or `%USERPROFILE%\.windbg-mcp\profiles.json`; raw `connection` still
+works for a target nothing is configured for, and is the last resort rather than the quick option.
+
+A raw `connection` now also reaches a second place: with recording on (below) it is written to the
+server's own transcript file, scrubbed to `key=<redacted>`. That backstop is not a reason to pass
+one. A profile keeps the key out of the request, so there is nothing for either transcript to
+redact, and redaction is a thing that has to keep working while a key never sent cannot leak.
+
+The live smoke tier below is the one sanctioned exception: `WINDBG_MCP_SMOKE_KERNEL` is a raw
+connection string, passed straight to `attach_kernel { "connection": … }`, and is deliberately
+*not* a profile — the tier has to exercise the explicit path, and it is a variable in a developer's
+own shell rather than something a client ever sees.
 
 A worker process does **not** inherit `WINDBG_MCP_PROFILE_*` (`engine::spawn_worker` strips them):
 it is told the one connection it is opening over its private pipe, and a `launch`ed debuggee would
@@ -224,6 +231,12 @@ Two things worth knowing when using it here:
 - **`windbg-mcp --render-cast <transcript.jsonl>`** turns one into an asciicast. That is the
   supported way to produce the recordings under `examples/` and `docs/` — the older ones are
   hand-reconstructed and say so, and a new walkthrough should not add another.
+
+Recording a **live kernel** session is where this needs care, because a transcript of one is as
+sensitive as the target: not the connection (attach by `profile` and there is no key in it), but
+everything the debugger printed — stack frames, strings, whatever the guest holds. Nothing but
+secrets is masked, so treat the file like a crash dump: keep it out of the repo, and delete it when
+the investigation is done. It is **appended** to, so a path reused across runs accumulates.
 
 ## Plugin vs. dev build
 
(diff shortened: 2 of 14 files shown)


glslang:
@coderabbitai review

glslang:
Addressed in d62bb82. All sixteen findings triaged; thirteen were distinct real issues (several were reported by both reviewers), and all thirteen are fixed.

> [!NOTE]
> **A correction to that commit message**, which claims "Ten were real; each was reproduced before it was fixed." Both halves are overstated. Thirteen were real, not ten. And four were reproduced before fixing — the two scrub holes with a probe, and the multi-run cast and the null session with negative controls against the new tests. The other nine were established by reading the code, and are pinned by tests going forward, but I did not watch them fail first.

### Redaction — two holes, in opposite directions

- **`{"password" : "hunter2"}` was recorded verbatim.** The scan required the separator immediately after the name. Filler is now allowed on both sides, which also closes the case that matters more: `net:port=1, key = 1.2.3.4` is refused by `is_dialable` — but only *after* the argument has been recorded, so a scan insisting on `key=` exactly masked the key in the connections this server accepts and missed it in the one it turns away.
- **The opposite mistake, same lines:** `trim_start` skipped newlines, so `key=` at the end of a line masked the first token of the *next* line. Filler is now spaces and tabs only. Prose stays safe because that rests on the quoted-`:` rule, not the whitespace rule — `Key: \REGISTRY\…` is untouched.
- **The interrupt detail** was the one path recording debugger text without scrubbing it. Now scrubbed and capped like every other.

### Events that were not true

- **A disconnect recorded `shutdown` and then nothing per session** — the one teardown with no caller to answer, and the transcript never said whether each target was released or a worker killed still holding one. It now writes the same `session_end` `end_session` does, with `released` read from `ShutdownNote` so a teardown that lost a race doesn't report the target as still attached.
- **`worker_lost` fired on orderly teardowns.** `release` terminates the worker on purpose, so EOF there is the last step of a successful `end_session`. Emitted only when the death is news.
- **MRTR `InputRequired` was recorded as a completed empty result.** Nothing is recorded now; the request record standing alone is what actually happened.

### Ordering and bounds

- **`seq`/`mono_ms` were allocated before the file lock**, so two writers could be numbered in one order and write in the other. Lock first — the two orders are now one order by construction.
- **Appended runs made the cast step backwards.** Each run's `mono_ms` restarts at zero; rendered as they stood the join went `0.03 → 0.0` (measured against the old code). Runs are laid end to end, separated by the wall-clock gap, and ordered within a run by `seq` so a file from before the lock fix still plays.
- **Derived batch records bypassed the cap**, and the step label — the caller's own string — was not capped at all. Both now take `WINDBG_MCP_TRANSCRIPT_MAX_FIELD`; `step` became a `Capped`.
- **A payload that will not serialise** was measured as zero and kept, which then discarded the whole record in `write`. Dropped to a marker instead.

### Correctness and cost

- **`session: null` on every default-routed call.** Omitting `session_id` accepts the current session, not none — so with two targets open the transcript could not say which was read or changed. `run_call` reports where it routed via a task-local scoped to the call; openers report the session they minted. Threading it back would have changed forty-odd tool signatures to carry a fact none of them uses.
- **`debug_batch`'s two preflight refusals** answered with text alone despite the new output schema. Now the typed `status: "error"` branch with an `invalid_argument` category.
- **`text_of` ran even with recording off**, joining and copying every content block of every result. The whole hook is skipped when the recorder is disabled.

### The test that could not fail

`nothing_is_recorded_unless_a_transcript_is_asked_for` asserted a path the server was never given, so no default could have written it. It now uses the same path twice — once with the variable, proving the server writes that exact file, once without.

### Not taken

Sorting *all* records by `seq` in the renderer. Across a multi-run file that is wrong, because each run numbers from zero; the sort is per run, inside the segmentation the timeline needed anyway.

---

New coverage: five tests, three of them checked against the old code first — the two scrub cases, a multi-run cast, an out-of-order file, the derived-field caps, plus smoke tests for both teardowns, for default routing with two sessions open, and the rewritten opt-in test. `cargo test` 360 unit + 44 smoke, clippy and fmt clean, debugger tier green.

glslang:
Addressed in b0d421d. All five fixed — and one of them was a bug in the previous round's fix, which is the one worth reading.

### The reorder that only restamped

The multi-run fix sorted each run's records by `seq` to compute their times, then wrote those times into **file-order** slots, and `render` still zipped them with the records in file order. So an out-of-order file came out with rising timestamps attached to the wrong records — a result played before the request that produced it. A recording of something that did not happen is worse than one that will not play.

**My test passed**, because it asserted the times rose. That is the defect itself, restated as an assertion. `timeline` now returns records in playback order rather than a parallel array of times, `render` iterates that order, and the test asserts the *content* order (the call before its answer) with monotonic time as the secondary claim. Checked against the old behaviour.

### Two servers sharing a path

`to_file` opens for append, and the docs said in as many words that two servers may point at one file — while claiming their runs are "told apart by the `start` record and its pid". That holds only for runs that do not overlap. Concurrent supervisors interleave their lines, and everything after the later `start` was read as one run, mixing two sets of sequence numbers, request ids and sessions each numbered from scratch.

Every record now carries a `run` id — wall clock mixed with the pid, so neither a shared nanosecond nor a reused pid can collide — the renderer groups by it, and the module docs no longer claim an interleave cannot happen. It can; this is what makes the file readable when it does. New test renders two recorders writing alternately to one path and asserts each server's call and answer come out adjacent.

### Three narrower ones

- **A `session_id` carrying a credential was recorded verbatim.** The argument object is scrubbed; the copy lifted out of it for the `session` field was not — and on a handle that does not resolve, nothing replaces it, so it reached the result record too. `kdconn` already refuses a connection string typed into `profile` *without echoing it back*, and a transcript writing it to a file would undo exactly that care. Good catch on the P1.
- **A quoted value containing an escaped quote leaked its tail.** `{"password":"hunter\"suffix"}` masked `hunter\` and left `suffix` — a line that reads as redacted while still carrying the secret, which is the worst shape a redaction bug takes. Now finds the first *unescaped* quote. My previous comment there ("none of these values contains an escaped one") was an assumption, not a fact; a `.server` password can contain anything.
- **`SessionOpen.target` was uncapped.** A `launch` target is a whole command line and it is the caller's, so with a small configured cap it produced a record over the documented limit with no marker. It is a `Capped` like everything else now.

---

`cargo test` 363 unit + 44 smoke, clippy and fmt clean, debugger tier green, and re-driven end to end against the sample dump.

glslang:
Addressed in 4bc8a37. All four fixed — and the shape of the first two is worth naming: **each round of review has found another *place* that forgot to record something.** Rather than adding the third and fourth call site, both are now fixed at the one point every path goes through.

### Teardowns, recorded in `release`

A session reclaimed to pay for a new one at the limit recorded nothing: `reconcile_capacity` releases in a background task with no caller to answer and nothing in any tool result to say it happened. That is the third teardown, after `end_session` and the disconnect — and the first two rounds of review each found one of the *other* two missing. Three for three is a pattern, not three bugs.

The record now lives in `release`, which all three call. A caller cannot forget it, and a fourth teardown gets it for free. Two details worth flagging:

- `released` is the session's own flag rather than this attempt's outcome. When two teardowns race for one session only the winner is told it worked, and the loser's failure is not news about the *target* — "was it let go?" is what a transcript is read for, and the flag answers it. This can differ from `end_session`'s own structured result, which correctly reports *that caller's* attempt; both are true and they answer different questions.
- A stale handle still records nothing, because it returns before this point — no teardown that tore nothing down appears as one.

### Routing, recorded in `resolve`

`interrupt` and `end_session` resolve by hand instead of going through `run_call`, so last round's routing fix never ran for either. Called without `session_id` — the common way to call `end_session` — the record said `session: null` for a call that had just torn a target down. Noting the route moved into `Sessions::resolve`, the funnel all three take; the task-local moved to `record` so the engine can reach it.

Openers mint rather than resolve, so they still say so explicitly — but now from **every** outcome carrying an id. `PostCommit` left a target open and `Timeout` left an open that may still land; both are the recovery cases where the next question is "what happened to that session?", and the answer is in events a null field cannot be joined to. `opened_session` is a `match`, so a new `OpenError` variant is a compile error rather than an outcome that quietly stops being recorded.

### The renderer could destroy its own input

`--out` naming the transcript, or a transcript already called `.cast` (whose default output is itself), and `File::create` truncates the source *after* it has been read — so the render reports success and replaces the only JSONL copy of the session with a rendering of it. That is the one outcome this whole feature exists to prevent, arriving through the tool meant to make a session shareable. Refused now, before a byte is read, comparing by `canonicalize` where the filesystem can answer.

---

Each fix has a test that fails without it. The reclamation one I checked against a build with the record suppressed, since it is the P1 and it depends on a background task landing. 364 unit + 45 smoke, clippy and fmt clean, debugger tier green.

glslang:
Addressed in 2d5c7e9. **Two of the five were the third instance of the same complaint**, so I stopped patching instances and closed the categories instead.

| Round | "a field is not capped" | "the scan misses a syntax" |
|---|---|---|
| 2 | derived batch details | whitespace around the separator; a line break after it |
| 3 | `SessionOpen.target` | an escaped quote inside a value |
| 4 | the caller's `session_id` | a backslash-escaped member name |

Three for three, both columns. Fixing the fourth instance of each would have invited a fifth.

### Capping: bound the record, not each field

Every instance was a *new place that had to remember*. `write` now bounds the record itself: anything past a ceiling far above what a capped record reaches is replaced by an `oversized` marker naming the kind and the size. A field nobody capped — including one added later — can no longer put an unbounded line in the file, and it surfaces as a fact to go and fix rather than as a huge line nobody notices. The `session_id` instance is capped too, since that was the specific report.

### Redaction: mask by value, and demote the scan to a net

The scan has to anticipate how a secret is *written*, and it had been wrong four times in four ways with no reason to think the list was finished. `scrub` now masks **by value** first: every secret this server is handed is remembered at `Connection::new` — the one door a profile and a raw argument both pass through — and masked wherever it occurs, in any syntax. A key that reached this process cannot leave it in a transcript however the surrounding text is written.

The pattern scan stays as a net for a secret this server has never seen (one a target printed itself), is now documented as best-effort rather than as the guarantee, and also learned the escaped-name shape. Values under six characters are not masked by value — a key of `1` would blank every digit in a transcript, which is worse than the leak.

One consequence worth flagging: the registry is process-wide by design, so a test asserting something is *not* masked cannot reuse a value another test has dialled. One did; it now uses its own.

### Three that were their own thing

- **A hostile target could run commands on a viewer's terminal.** A cast frame is bytes a player writes straight to a terminal, and transcript text is whatever the target produced — and these targets are dumps and drivers chosen for analysis, played back by whoever the recording was sent to. OSC 52 writes the viewer's clipboard; others retitle the window or make the terminal type its answer back as input. Everything from the transcript is now rendered visible (`^[`); the renderer's own styling, added from constants outside that, still works. Good find — this was the most serious thing in the round.
- **`session_open` was recorded before `admit` accepted the session**, so a disconnect racing a handshake could put an open *after* the shutdown record for a session that was never registered and has no end. Recorded after admission now; a worker refused a slot holds no target and is recorded as nothing, which is what happened.
- **The README example could not be produced by this code** — no `run`, `target` shown as a string rather than the `Capped` object. The three lines are now copied from a real transcript rather than written by hand, which is why they were wrong.

---

369 unit + 45 smoke, clippy and fmt clean, debugger tier green.

Issue #120, opened 2026-08-13
modules: render the listing from its own values, not lm's text


glslang:
## Problem

`modules` is the one tool whose **text is the debugger's rendering rather than a rendering of its own values**. `structured.rs` states the rule the rest of the server follows — the text is a rendering, `data` is the same answer as values, and a figure re-derived from a rendering measures the rendering ([#77](https://github.com/glslang/windbg-mcp/issues/77)) — but here the two halves are produced independently: the text comes from `lm` / `lm m <pattern>` inside DbgEng, and the values from `IDebugSymbols3` records matched in this process.

Every other tool that emits both already renders its text from its values: `crash_triage`, `pool_census`/`pool_diagnostics`, `walk_memory`, `set_breakpoint`, `run_to_address`. `modules` pastes `lm`'s output.

That seam is not theoretical. Adding a `filter` in [#119](https://github.com/glslang/windbg-mcp/pull/119) meant implementing WinDbg's string wildcard grammar a second time, and it took four review rounds — each a real case where the same call's text and values described different sets of modules:

| what diverged | measured on the sample dump |
|---|---|
| character sets | `lm m nt[fd]*` prints `Ntfs`; matched literally, nothing |
| the `\` escape | `lm m n\t*` prints `nt`, `Ntfs`, `ntosext`; literally, nothing |
| whitespace | `lm m nt v` matches `nt` and prints lm's **verbose** listing — a different command, not a different filter |
| case folding | the matcher folds ASCII; DbgEng folds by Windows' rules, so `é` could match in the text and miss in the values |
| unloaded modules | `lm` appends its `Unloaded modules:` tail to a filtered listing; the values carried loaded ones only (fixed by [win-kexp#101](https://github.com/glslang/win-kexp/pull/101)) |

Four of those five are now handled by **refusing input** — an allowlist of "an ASCII name plus `*` and `?`" — which is the honest fix while two implementations exist, and is a strange thing to have to tell a caller about a listing tool. None of them can happen if there is only one implementation.

## Proposed shape

**The enumeration already exists**, which is what makes this a subtraction rather than new engine work. `IDebugSymbols3` returns the modules — `GetNumberModules`, `GetModuleParameters` over the loaded range and the unloaded one, `GetModuleNames` — and win-kexp already wraps both halves as `modules()` and `unloaded_modules()`. This tool calls them today and then *also* runs `lm` purely to have something to print. Filtering is a local operation over records already in hand; the command is the only part that has to go.

[...]

PR #156, opened 2026-08-18
Give a stack frame a coordinate, not just a symbol

glslang:
`backtrace` ran `k` and handed back whatever DbgEng printed. That names a frame as
`module!Symbol+0x1c`, which is unusable the moment the symbol does not resolve — and on a driver
with no PDB it never does. `crash_triage` has answered `{module, rva, symbol}` per frame since it
was written, for exactly that reason; this is the same answer from the tool people actually call
for a stack.

This is phase 2 of the split-plane plan ("coordinates"), first half. It is what makes any analysis
server pair with the debugger: a frame keyed by `(module, rva)` joins a disassembler's function
list without either side knowing the other exists.

## The frames are literally the same records

Both tools walk through one helper (`worker::walk_attributed`) and render through one function
(`triage::describe`), so a frame carried from one to the other names the same place by
construction rather than by two implementations agreeing on inspection. The debugger tier asserts
it from outside, comparing the two stacks field by field with `analyze: false` so neither call
moves the scope.

## Three things this had to decide

**A cap, which `k` did not have.** 32 frames by default, 256 at most, with `frames_truncated`
saying when the stack went on — the same walk-one-extra trick `crash_triage` uses, so a stack that
is exactly 32 long is not reported as cut off. The frames are values now, and an uncapped typed
answer is an uncapped bill for whoever reads it. Twice `crash_triage`'s depth because its own cap
comment points here for the deep stack.

**The listing is rendered from the values**, as `modules` is (#120), so the text and the records
cannot disagree. The cost is stated rather than hidden: this is not `k`'s output — no
`Child-SP`/`RetAddr` columns, and no `[Inline Frame]` rows, which a stack walk does not return at
all. `execute { "command": "k" }` is that listing verbatim.

**The walk failing fails the call**, unlike in a triage. A triage that cannot walk still has a bug
check to report; here the stack *is* the answer, and an empty one would claim this thread has no
frames when what happened was a read failing. An empty walk that really is empty says so in prose.

## What it costs

Under the client behaviour measured in `docs/token-budget.md`, `structuredContent` **replaces**
the text block, so a model driving this now reads the frames and not the listing. That is the
point, and it is the same trade `FOLLOWUPS.md` item 24 records for every other typed tool.

| | before | after |
|---|---:|---:|
| surface, model-visible | 495 B | 1,134 B |
| surface, wire | 578 B | 8,156 B |
| result, model-visible | 572 B | 945 B |
| result, wire | 661 B | 1,588 B |

The surface wire jump is the inlined `$defs` finding 1 of item 24, not something this adds; the
result numbers are against different dumps and `docs/token-budget.md` says so beside the row.
`backtrace` leaves that item's uncapped list. The result ceiling is 3,000/4,000 — looser than 45%,
on the same reasoning `crash_triage` is: a stack whose frames resolve carries a `symbol` on every
one where an offline walk carries none.

## Testing

On the ARM64 bench (`aarch64-pc-windows-msvc`), against the native sample dump:

- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --bins` — 436 passed, including 5 new rendering tests
- `WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke -- --test-threads=1` — **60 passed, 0
  failed, 12 ignored**, run twice. The dump tier's new assertions ran rather than standing down:
  no `SKIPPED` on this host, so the `crash_triage`/`backtrace` join was really compared.
- Goldens re-recorded with `UPDATE_GOLDEN=1`; the diff is `backtrace` gaining an output schema and
  a `frames` param, and `crash_triage`'s output schema growing 202 B from a shared doc comment.

One flake seen and not reproduced: `ending_a_session_stops_a_running_batch_and_rolls_it_back`
failed once in a full serialized run ("the rollback finished and the client was never told") and
passed on re-run, on a full run of `main` in the same configuration, and in isolation on both
branches. Unrelated to this change.

## Follow-ups

- Phase 2's other half is `disassemble`, then image identity (PDB GUID + age) on `ModuleInfo`,
  which needs a `win-kexp` primitive.
- The phase's acceptance test — joining a frame to a Binary Ninja function by hand — needs a
  driver image on the analysis machine and has not been run.

## Summary by CodeRabbit

- **New Features**
  - `backtrace` now returns structured stack frames with symbols, module names and relative virtual addresses.
  - Added configurable frame limits, defaulting to 32 and capped at 256.
  - Results indicate when the stack trace has been truncated.
  - Structured output is now shared consistently with crash triage.

- **Documentation**
  - Updated tool documentation, changelog and output-size guidance to describe typed frames, limits and rendering differences.


glslang:
@coderabbitai review

glslang:
Addressed in 074dde2. All four findings were valid; here is what each turned into.

**Codex P1 — playbooks still equating `backtrace` with `k`.** Fixed, and it was worth more than a
label change: `skills/windbg-debugging/crash-dump.md` step 5 also told readers to check for
`module!name` to decide whether they have symbols, which is now the wrong test — `module`+`rva` is
there either way, and an absent `symbol` is the signal. It now says that, gives the cap and
`frames_truncated`, and points at `execute { "command": "k" }` for `Child-SP`/`RetAddr` and inline
frames. `examples/drive_kernel_test.ps1`'s label and `skills/windbg-debugging/SKILL.md`'s tool
table updated too.

**CodeRabbit, `src/worker.rs` (Major) — unescaped frame lines inside a fence.** Correct, and this
was a reintroduction rather than a new gap: `renderable()` exists precisely for it, and #126 fixed
the same class for module names. A frame's `module` and `symbol` come from the target, so a module
named with three backticks closed the block and injected Markdown. The whole rendered line now goes
through `renderable()` — the line rather than the fields, so a coordinate added later cannot escape
by being added in the wrong place — and
`worker::tests::a_frame_name_cannot_close_the_fence_it_is_printed_in` uses the same hostile name as
the module-listing test beside it.

**CodeRabbit, `CHANGELOG.md`/`README.md` — the contract was over-promised.** Also correct. "Every
frame now carries the offset into its image" is false for a frame the engine places in no loaded
module, which is exactly what a driver bug leaves behind. The CHANGELOG now says `module`/`rva`
travel together and can both be absent, `symbol` likewise, and that `address` is the only field
always present — plus that an RVA is comparable across machines only against the same image **and
build**, which is the half of this phase that has not landed. `backtrace` has a row in README's
Structured results table, and `crash_triage`'s row points at it rather than repeating the caveats.

**CodeRabbit, `tests/golden/tool_budget.json` — doc totals out of sync with the golden.** Correct;
I updated the results row and missed the surface totals. `docs/token-budget.md` is re-stated from
this run: payload 366,431, sum 366,313, model-visible 66,717, and the `outputSchema` line now says
32 tools rather than 31. The stale baselines in the two ceiling doc comments in `tests/mcp_smoke.rs`
went with them.

One note on the two-revision table, since it is the one number a golden does not own: the
`2025-06-18` payload is **measured**, not derived from the old 56-byte gap. Driving the server at
both revisions and measuring the serialized `result` reproduces the `2026-07-28` figure to the byte
against the golden (366,431), which is what makes the other figure (366,375) worth writing down.
The first attempt at that measurement was wrong by ~1.5 KB because the console decoded the
descriptions' non-ASCII through the default code page — worth knowing if anyone measures this again.

Re-verified on the ARM64 bench: clippy clean, `cargo test --bins` 437 passed (one new), and
`WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke -- --test-threads=1` 60 passed / 0 failed /
12 ignored. CI is green on all ten checks.

glslang:
Codex's P2 on `074dde2` is addressed in the latest commit — I'd replied before it landed, so it was
not covered above.

**It is right, and it lands on the one distinction this code was already careful about.**
`Attribution` has three states precisely because two of them are opposite facts that both arrive as
an absent module: "no loaded module holds this address" is a *positive finding*, and one a driver
bug produces constantly — a freed pool page, an unloaded driver, a corrupted return address — while
"the lookup failed" is an absence of information. `is_candidate_culprit` keeps them apart, and
`triage.rs`'s own doc comment says at length why collapsing them is wrong.

`frame_info` then collapsed them anyway. Both serialized as no `module` and no `rva`, so a consumer
reading a frame with no coordinate could not tell which it had — and would reasonably read the
failure as the finding, which is a claim about the target made out of a call that did not answer.
Survivable while the frames were a triage's supporting evidence and its own heuristic saw the enum;
not survivable now that the coordinate is a tool's whole answer.

`FrameInfo` therefore grows `attribution_failed`, skipped when false — every frame of an ordinary
walk — so the normal path costs nothing, and the model-visible surface does not move at all
(66,717 B either way; the field is in an output schema, which no model reads). `describe` says it
on the line too, because the misreading is one a person makes at a glance: a bare address on a
kernel stack looks like a freed pool page.

Not the other option Codex offered — failing the call. The remaining frames are perfectly good, and
a walk that gives up because one lookup failed answers with less than it knows; triage has always
treated attribution as best-effort for that reason.

`crash_triage` gets the field too, the frames being the same records. That is the shape of this
whole change working as intended rather than a side effect.

Costs, re-recorded: output schemas +1,620 B on the wire (payload 366,431 → 368,051), model-visible
unchanged. `docs/token-budget.md` and the two-revision table are re-stated again from this run, the
`2025-06-18` figure re-measured (367,995) rather than carried over.

Verified on the ARM64 bench: clippy clean, `cargo test --bins` 439 passed (two new — one at the
`frame_info` level, one at the rendering level), and the full dump tier 60 passed / 0 failed / 12
ignored. All ten CI checks green.

PR #157, opened 2026-08-18
Give an instruction a coordinate, and an encoding to check it against

glslang:
`win-kexp` is pinned at [`c20c695`](https://github.com/glslang/win-kexp/commit/c20c69588a3335b747e78fe0745628c96413cdcc)
— the merge of [glslang/win-kexp#112](https://github.com/glslang/win-kexp/pull/112), which adds the
primitive this uses. Re-verified against that pin rather than against the branch it was developed on.

The second half of the coordinate work (phase 2), and the last tool that could only answer "what is
the code here" as a rendering. `disassemble` ran `u` and handed back one string: addresses in the
debugger's backtick form, columns aligned by padding. Reading the address of the third instruction
meant parsing it back out; pairing any of it with a disassembler meant parsing it back out and then
rebasing it.

Each instruction is now `{address, rva, bytes, text}`. The addresses are the engine's own
arithmetic rather than a re-parse — the win-kexp primitive walks `IDebugControl::Disassemble`,
whose every instruction reports where the next one starts — and a stack frame's RVA and an
instruction's are now the same number, written by the same function.

## Four things this had to decide

**The module is named once.** A disassembly is a *contiguous* range, so it is almost always inside
one image, and repeating that name on every row is the per-row scaffolding `docs/token-budget.md`
measures elsewhere. An instruction carries its own `module` only where a range has run off the end
of one image — the only case the hoist would otherwise get wrong.

**`bytes` is what identifies a build**, which makes the plan's "stale image in the analyser" risk
checkable: two images that disassemble differently are different builds whatever their names say.
It is the engine's spelling rather than memory order (`d503237f` on ARM64 is the instruction word),
so it compares against another disassembly, not against a file. Said in the field's own docs,
because the other reading is the tempting one.

**`stopped_early` is the code ending, not the call being cut.** Disassembly runs forward into
whatever follows a function, which may be unmapped or not code, so a short answer is a fact about
the target and asking again with a larger count returns the same instructions. A count the caller
chose is not that and does not set it — there is a test for that confusion.

**The debugger's backtick address form comes out of the operands.** An operand is full of them, this
server spells an address one way, and that tick is also the delimiter of the code span the listing
prints in — so left alone, every operand carrying an address rendered as `\u{60}` once the line was
made safe for its fence (found by looking at real output, not by reasoning). Removed only where it
is punctuation between two eight-digit halves: MSVC decorates real symbols with backticks
(`` `anonymous namespace' ``, `` `vftable' ``) and stripping those would quietly rename them.

## The coordinate now joins three tools

`disassemble` with no address starts at the program counter, which is where `backtrace`'s frame 0
is. The debugger tier asserts the two report the same `address`, the same `rva` and the same
`module` for it — alongside the existing `crash_triage`/`backtrace` frame comparison. Three tools,
one coordinate, checked from outside rather than by construction.

## What it costs

| | before | after |
|---|---:|---:|
| surface, model-visible | 747 B | 1,405 B |
| surface, wire | 811 B | 8,465 B |
| result, model-visible | 1,291 B (text) | 1,808 B |
| result, wire | — (no budget) | 3,219 B |

`disassemble` had no result budget at all before; it has one now (4,000/6,000, loose for the same
reason `crash_triage`'s is — operands change shape when symbols resolve). It also leaves
`FOLLOWUPS.md` item 24's uncapped list, which is now down to raw-text tools only. The surface wire
jump is the inlined-`$defs` finding, not new. `docs/token-budget.md` is re-stated from this run,
including the two-revision payload table, whose `2025-06-18` figure is measured rather than derived.

## Testing

ARM64 bench (`aarch64-pc-windows-msvc`):

- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --bins` — **445 passed**, including 8 new (split, backtick normalisation, the
  coordinate column, the empty and early-stop renderings, and a fence-escaping test at this third
  call site)
- `WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke -- --test-threads=1` — **60 passed, 0
  failed, 12 ignored**. No `SKIPPED` on this host, so the join assertions really ran.
- Driven by hand against the ARM64 sample dump first, which is how the backtick problem and a
  misleading coordinate column were found before any assertion was written around them.

win-kexp #112 is green on both `windows-latest` and `windows-11-arm`.

## Summary by CodeRabbit

- **New Features**
  - `disassemble` now returns structured results with instruction addresses, module-relative offsets, encoded bytes and readable instruction text.
  - Specify how many instructions to retrieve, from 1 to 128, with a default of 16.
  - Results indicate when decoding stops early and include the resolved starting address and module details.
  - Address operands are normalised for clearer output.

- **Documentation**
  - Updated usage guidance, crash-dump workflows and output-size information to reflect the enhanced disassembly results.


glslang:
@coderabbitai review

glslang:
`win-kexp` is merged and the pin now points at
[`c20c695`](https://github.com/glslang/win-kexp/commit/c20c69588a3335b747e78fe0745628c96413cdcc) on
`main`; everything below was re-verified against that pin rather than the branch it was developed on.

All seven review findings were valid, and the first two compound into a design bug that is now
fixed rather than patched.

**Codex P1 — a failed lookup collapsed into "in no module".** The same finding #156 fixed for a
stack frame, at the tool that computes the coordinate a second time. Correct, and following it
exposed something worse: with the module hoisted to the answer and per-instruction `module` meaning
"different from the range's", an instruction in **no** module serialized identically to one in the
range's module — so it was silently credited to the image around it. An absent field that means
"look at the range" cannot also mean "there is nothing to look at". **The hoist is gone**: every
instruction names its own image, as a frame does, and `attribution_failed` marks the lookup that
did not answer. Two tests pin the rows that were wrong.

**Codex P2 — the smoke assertion rejected every backtick.** Also correct, and it would have made
the tier symbol-dependent: `plain_addresses` deliberately keeps MSVC's decorated names
(`` `anonymous namespace' ``), and a host that resolved one would have failed an assertion about
address normalisation. It now tests for the address form specifically — eight hex digits, a tick,
eight more.

**CodeRabbit — `tool_error` where the tool now declares an output schema.** Right, and the sibling
tools it named all use `typed_error`. Fixed, with a test for that path: the coverage test only
exercises each tool's *session* refusal, so this branch had none.

**CodeRabbit — `saturating_sub` for the RVA.** Right: `held` reaches that line by a route that has
not been through the containment test, and `triage::frame_info` guards the same subtraction with a
comment naming the hazard. Now guarded the same way.

**CodeRabbit — the summary sentence quotes an unescaped module name.** Right, and the repo has
already fixed the same construct in `summary_text`. The rows were protected by the fence and by
`renderable`; the sentence is outside it. Escaped, with a test.

**CodeRabbit — docs treat `rva` as unconditional**, and **CodeRabbit — no test for the
per-instruction module.** Both fixed by the redesign above: README, CHANGELOG and the skill
playbook now state the omission rule, and the uncovered branches have tests because they are the
branches that were wrong.

Re-verified on the ARM64 bench against the merged pin: clippy clean, `cargo test --bins` **449
passed** (four more than before), `WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke --
--test-threads=1` **61 passed / 0 failed / 12 ignored**. Goldens re-recorded and
`docs/token-budget.md` re-stated from this run, `2025-06-18` re-measured. All ten CI checks green.

glslang:
Two more Codex P2s arrived on the previous commit, both leftovers of the redesign above — the hoist
was removed from the data but not from everything that described it. Fixed in the latest push.

**The summary sentence still said "in no loaded module" when the lookup had failed.** The *row*
already said `module lookup failed`, so one answer carried both readings two lines apart — which is
precisely the conflation `attribution_failed` exists to end. The sentence now says the engine could
not name the module, with a test that the two never disagree again.

**The tool description still said the range's module is named once.** That is the model-visible
half, so it was the worst place for it to be stale: a client following it would inherit the first
instruction's module across a range that crossed into another image, or across an instruction in no
module at all — the exact bug the redesign removed from the data. It now says to read the pair per
instruction and why.

`cargo test --bins` **450 passed**, full dump tier **61 passed / 0 failed / 12 ignored**, clippy
clean, all ten CI checks green. Goldens re-recorded and `docs/token-budget.md` re-stated once more
(the description change moved the model-visible total to 67,613 B), with `2025-06-18` re-measured
against this run.

No findings outstanding.

PR #158, opened 2026-08-18
Say which PDB the engine has, so a client elsewhere can fetch that one

glslang:
`win-kexp` is merged and the pin points at
[`ef8e4cb`](https://github.com/glslang/win-kexp/commit/ef8e4cb3e29c7314bce1bf34655068388dcb36c2) on
`main`. Re-verified against that pin rather than the branch it was developed on.

The last piece of phase 2's coordinate work, and it comes with the doc that motivated it.

An image was already identified: `timestamp` + `size` on every `modules` row is the pair a symbol
server is keyed by. [`docs/coordinates.md`](docs/coordinates.md) — new here — walks that end to
end, with real numbers: from those two integers a Mac that had never seen the target fetched a
byte-identical `ntkrnlmp.exe`, confirmed both fields against the PE header, and resolved a
`crash_triage` frame's RVA to `KeBugCheckEx` in the export table at exactly the entry its
displacement predicted. All four instruction encodings `disassemble` reported matched the file
byte-for-byte. **No proxy, no upstream call** — which is what section 06 of the plan predicted.

What that exercise found is that only *exported* frames resolve from the image alone. The internal
ones need the PDB, which is keyed by a different pair — GUID + age — that nothing reported. This
adds it.

## What `modules` now carries

```json
{ "name": "nt", "symbols": "pdb",
  "pdb": { "guid": "FE3F58BDA39D2FC13C370618D1DBDF22", "age": 1,
           "key": "FE3F58BDA39D2FC13C370618D1DBDF221" } }
```

- **`key` is carried already-built** — the middle element of `<pdb>/<key>/<pdb>` — because the age
  goes in **hex**, and getting that wrong produces a URL that 404s.
- **Absent when `symbols` is `deferred`**, which on a freshly opened dump is nearly all modules,
  and asked for only where the engine has already said there is one: a listing is two hundred rows
  and this is a call into the engine per row. The field means "the PDB this engine has", not "the
  PDB that exists".
- **`unmatched`** means the engine loaded a PDB that does not belong to the image, so every symbol
  on that module is another build's names.

**Scoped honestly**: this saves a download rather than enabling something otherwise impossible. The
same identity is in the image's own debug directory, so a client that fetched the image can read it
there. What it adds is skipping that fetch, and being able to check a PDB you already hold is the
right one — which the image cannot do. The acceptance test established that ordering before this
was written, and the doc says so.

## Two things worth keeping from building it

**Do not copy a not-found mapping between calls.** `module_at` treats `E_INVALIDARG` as "no module
holds this offset" because that is measurably what the engine means there; reusing it here would
have turned a broken call into a believable claim about the target.

**One small type cost 15,610 B of wire and no model context at all.** `ModuleInfo` is embedded in
the openers' summary, in `modules` and in the allocator shapes, and `schemars` inlines `PdbInfo`
into every one. That is `FOLLOWUPS.md` item 24's first finding with a price on it. `WIRE_CEILING`
moves 412,000 → 460,000 with the reason recorded in `docs/token-budget.md` rather than absorbed
quietly, because that number moving is the only warning the duplication is compounding.

This PR also corrects `disassemble`'s row in `docs/token-budget.md`, which was measured before
#157's own review redesigned it and has been 210 B light ever since.

## Testing

ARM64 bench: clippy clean, `cargo test --bins` **450 passed**,
`WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke -- --test-threads=1` **61 passed / 0 failed /
12 ignored**. The debugger tier asserts the GUID's spelling and that `key` is the GUID with the age
in hex — the composition whose failure mode is a 404.

Verified by hand against the sample dump: `nt` reports the same GUID and age that `!lmi nt` prints
and that the engine's own symbol cache path uses, which is the key that fetched the correct PDB in
the doc.

## Summary by CodeRabbit

- **New Features**
  - Module listings now include PDB identity details for resolved symbols, including GUID, age, symbol-server key, and unmatched status.
  - Symbol metadata remains omitted when symbol resolution is deferred.

- **Documentation**
  - Added guidance on portable module coordinates, symbol lookup, image verification, crash-frame correlation, and common address pitfalls.
  - Expanded module result documentation with timestamps and symbol information.

- **Chores**
  - Updated supporting diagnostics and output-size limits to accommodate the additional metadata.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **`modules` reports which PDB the engine has for a module** — `guid`, `age`, and the `key` those
+  two make, which is the middle element of a symbol server's `<pdb>/<key>/<pdb>` path. It completes
+  the coordinate work: the image was already identified by `timestamp` + `size`, and its *symbols*
+  are identified by a different pair that nothing reported.
+
+  - **`key` is carried already-built** rather than left to the caller, because the age goes in
+    **hex** and getting it wrong produces a URL that 404s — a hard failure to read backwards.
+  - **Absent for a module whose symbols are `deferred`**, which on a freshly opened dump is nearly
+    all of them. This reports the PDB the engine *has*, not one it could find; that is not the same
+    as "this module has no PDB", and the `symbols` state beside it says which.
+  - **`unmatched`** says the engine loaded a PDB that does not belong to the image, which makes
+    every symbol on that module another build's names. Absent when false.
+  - It saves a download rather than enabling something otherwise impossible — the same identity is
+    in the image's own debug directory, and [`docs/coordinates.md`](./docs/coordinates.md) walks
+    the whole chain, including the part where an 11 MB image is selected out of a symbol server by
+    two integers.
+
 - **`disassemble` answers with typed instructions carrying `RVA` and encoding, not just `u`'s
   text.** The second half of the same coordinate work, and the last tool that could only answer
   "what is the code here" as a rendering. Each instruction is now
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1079,7 +1079,7 @@ checksum = "72069c3113ab32ab29e5584db3c6ec55d416895e60715417b5b883a357c3e471"
 [[package]]
 name = "win-kexp"
 version = "0.1.0"
-source = "git+https://github.com/glslang/win-kexp?rev=c20c69588a3335b747e78fe0745628c96413cdcc#c20c69588a3335b747e78fe0745628c96413cdcc"
+source = "git+https://github.com/glslang/win-kexp?rev=ef8e4cb3e29c7314bce1bf34655068388dcb36c2#ef8e4cb3e29c7314bce1bf34655068388dcb36c2"
 dependencies = [
  "byte-strings",
  "goblin",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.9.0"
 edition = "2024"
 
 [dependencies]
-win-kexp = { git = "https://github.com/glslang/win-kexp", rev = "c20c69588a3335b747e78fe0745628c96413cdcc" }
+win-kexp = { git = "https://github.com/glslang/win-kexp", rev = "ef8e4cb3e29c7314bce1bf34655068388dcb36c2" }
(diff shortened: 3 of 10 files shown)


glslang:
@coderabbitai review

glslang:
`win-kexp` is merged and the pin points at
[`ef8e4cb`](https://github.com/glslang/win-kexp/commit/ef8e4cb3e29c7314bce1bf34655068388dcb36c2);
re-verified against it. All five actionable findings were valid.

**Codex — the opener's `primary_module` skipped the enrichment.** The sharpest one, because it
broke a promise rather than omitting a field: `TargetSummary::primary_module` documents itself as
exactly the row `modules` returns, and enriching only the listing made that false — `open_dump`
and `modules` would have described the same module differently. Both now go through one
`with_pdb_identity`, so they cannot drift again.

**Codex — DIA-backed symbols were skipped.** Right: `SymbolKind::Dia` is the same PDB reached
through the Debug Interface Access API, and the predicate only matched `Pdb`, so a module in that
supported mode reported no identity while the engine held one. Both states enrich now.

**Codex and CodeRabbit both — gate the smoke assertion on `engine_resolves_kernel_symbols`.**
Correct, and it is the mistake that gate exists to prevent: reading a target and *resolving* its
symbols fail apart, and `pdb` is documented as absent for a deferred module — so on a host that
reads memory and resolves nothing, the assertion would have reported the documented contract as a
regression. It is gated now and prints `SKIPPED` with the shared reason.

**CodeRabbit — no offline test for the wire contract.** Right, and the smoke assertion could not
have caught the interesting case anyway: the contract has four ways to produce a 404 (the GUID
spelling, the age in hex not decimal, the concatenation order, `unmatched` appearing only when
true), and a real `nt` reports **age 1**, where hex and decimal are identical. The new test next to
`PdbInfo` uses age 26 — `0x1A` — so the confusion the field is most likely to suffer is the one
actually asserted.

**CodeRabbit — `unwrap_or_default()` masked missing fields.** Right: a missing `age` defaulted to
zero agrees with a key built from zero, so the schema regression would have passed. Read strictly
now.

The `Cargo.toml` comment asked for the Rust checks to be run; they were, and the pin has since
moved to the merged commit and been re-verified.

ARM64 bench: clippy clean, `cargo test --bins` **451 passed**, dump tier **61 passed / 0 failed /
12 ignored**. All ten CI checks green. Also corrected a `docs/token-budget.md` footnote I made
inconsistent — `modules`'s row is now measured on the same bench as the other two re-measured rows,
and says so.

PR #159, opened 2026-08-18
Hand the coordinate work over: the traps, not just the result

glslang:
Documentation only, and all of it is state a later session would otherwise rediscover the slow way
— the same shape as the ARM64 handover that preceded this line of work.

## `CLAUDE.md` — the four things that cost real time across #156–#158

None are in it today, and each was found by hitting it:

- **`cargo update -p win-kexp` does not move the pin.** `Cargo.toml` names an exact 40-character
  `rev`, so the update command re-resolves *that* revision and leaves you building the old code.
  This file said the command alone was enough.
- **Develop against the pushed feature branch, not a `[patch]`.** Pointing the `rev` at the branch
  commit needs no local checkout on the build machine and travels through git like everything else;
  repoint to the merge commit before the dependent PR merges.
- **The dev exe locks too, and quietly.** A worker left running holds
  `target\debug\windbg-mcp.exe`; `cargo build` then fails with `Access is denied (os error 5)`
  *after* compiling successfully, and a hand-driven run executes the old code — which reads as the
  change not working. It did, for one confusing round. Also noted: `cargo clippy` and
  `cargo test --bins` do not refresh that exe.
- **Both repos need `--admin` to merge.** A solo maintainer cannot self-approve, and this harness
  refuses that call, so an agent's job ends at "green and waiting" and the human merges.

Plus two smaller ones: redirecting the server's stderr from a driver script without draining it
deadlocks it as soon as `RUST_LOG` widens (it looks exactly like a hung debugger), and win-kexp's
clippy has four pre-existing ARM64-only errors — check against `main` before assuming they are
yours.

## `docs/smoke-test.md` — what the new assertions actually claim

The coordinate work added assertions whose *reasons* are not inferable from their names: the
three-way agreement between `crash_triage`, `backtrace` and `disassemble` (including why the triage
runs with `analyze: false` — `!analyze -v` moves the scope, so with it off the two stacks are the
same stack rather than two that ought to match), the shape claims that hold on a host resolving
nothing, and the PDB identity's own premise, which is symbol *resolution* rather than a target read
because `pdb` is documented as absent for a deferred module.

## `FOLLOWUPS.md` — the two things left behind

**Item 24's first finding has a price now.** One optional four-field type cost **15,610 B of wire
and zero model context**, because `schemars` inlines it into every schema embedding `ModuleInfo`.
That is the compounding the item predicts, measured on a change too small to have prompted anyone
to look, and `WIRE_CEILING` moved 412,000 → 460,000 to record it.

**Item 27 is new**: a deferred module reports no PDB identity. It carries the reason it was left
that way — the acceptance test measured that the field saves a download rather than enabling
anything — and the two cautions closing it would have to respect (minidump headers may be absent;
reading a header structure is not reconstructing an image, but the boundary belongs in the code).

## Testing

`markdownlint-cli2` over the same globs CI uses: **0 issues in 20 files**. No code changed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,8 +54,9 @@ To rebuild and load the new code without stopping the session:
    ```
    This builds the `win-kexp` revision pinned in `Cargo.lock` and writes a fresh
    `target\release\windbg-mcp.exe`. If this `windbg-mcp` change depends on a newly pushed
-   `win-kexp` commit, run `cargo update -p win-kexp` first and commit the resulting `Cargo.lock`
-   bump with the `windbg-mcp` change. The running server keeps executing the *old* code from the
+   `win-kexp` commit, move the pin first — edit the `rev` in `Cargo.toml`, then
+   `cargo update -p win-kexp` — and commit both with the `windbg-mcp` change (see below: the update
+   command alone does **not** move a `rev` pin). The running server keeps executing the *old* code from the
    renamed `.stale` file until its connection is recycled.
 3. **Load the new binary** by reconnecting the server: `/mcp` → reconnect `windbg` (or restart
    Claude Code). Only after this reconnect do the windbg tools run the new code.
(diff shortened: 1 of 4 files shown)


glslang:
Both findings fixed, and the first one was dangerous advice rather than merely imprecise.

**Codex P1 — `taskkill /IM` kills the release server too.** Right, and worse than the comment says:
this file documents elsewhere that a worker killed while holding a broken-in kernel leaves the
*guest* frozen and `prlctl exec` hanging on it. So the recipe I wrote — in the section whose whole
premise is "no session restart needed" — could have taken down a live debugging session and frozen
a target, from advice about a stale dev build. It now selects by executable path and touches only
what is under `target\debug`:

```pwsh
Get-Process windbg-mcp -ErrorAction SilentlyContinue |
  Where-Object { $_.Path -like '*\target\debug\*' } | Stop-Process -Force
```

Worth noting the command as written *worked* where I ran it, because no release server was
connected on that host — which is exactly how a recipe that is only conditionally safe gets
written down as if it were safe.

**Codex P2 — paginate the review recheck.** Right, and self-defeating in the way that matters: the
command exists to catch findings arriving after a reply, and on a PR with enough comments to span
pages the first page holds the *older* rounds. `--paginate` added.

`markdownlint-cli2` over CI's globs: 0 issues in 20 files. No code changed.

glslang:
Three more findings arrived on the previous commit — surfaced by the paginated recheck this PR
documents, which is a reasonable first outing for it. All three valid.

**The `never locked` contradiction.** The paragraph above my addition still claimed the dev profile
is never locked, directly against the paragraph explaining that a stale worker locks exactly that
exe. It now says what is actually true: `target/debug` is what the *registered release server*
never holds.

**The three-way sharing claim was wrong, and it undersold the test.** I wrote that all three tools
share one walk and one renderer. Only `crash_triage` and `backtrace` do — `disassemble` reaches the
same coordinate independently, with its own `module_at` calls, its own record and its own
rendering, sharing only the function that formats an RVA. That independence is precisely why the
third comparison is the one that can catch something, so the doc now says so instead of implying
the assertion is redundant.

**A backtick address broke a code span** in the sentence about backtick addresses breaking code
spans. `` `fffff801`3c677ef0` `` closes after `fffff801` and the rest reopens as markup;
`docs/coordinates.md` already used the double-backtick delimiter for the same string, and this one
does now too.

`markdownlint-cli2`: 0 issues in 20 files. CI green.

glslang:
Fixed — and this one turns the PR from docs-only into a one-line test move, because the alternative
was shipping a doc that describes coverage the code does not have.

**Codex P2 — the PDB assertion sat inside the target-read gate.** Correct. I put it there because
that block already existed, not because it needed anything from it: `modules` and the engine's
symbol bookkeeping answer the PDB identity without touching a target page. So on a host that
resolves `nt`'s symbols but cannot read its base, the check was skipped while being perfectly
runnable — and the doc I was shipping in the same PR claimed its premise was symbol resolution. The
block now sits outside the read gate with only `engine_resolves_kernel_symbols` in front of it, and
the doc says the two premises fail apart in both directions.

Worth noting how this was caught: the paginated recheck this PR documents surfaced it, on the
commit that added the pagination.

Re-verified on the ARM64 bench: clippy clean, `WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke
-- --test-threads=1` **61 passed / 0 failed / 12 ignored**. `markdownlint-cli2` 0 issues in 20
files. CI green.

glslang:
Fixed. **Codex P2 — `dia` counts as resolved.** Correct, and it is the gate's own failure mode
pointed the other way: `with_pdb_identity` treats `Pdb` and `Dia` alike, because `dia` is the same
PDB read through the Debug Interface Access API, so on a DIA-backed host the assertion skipped a
check it could run — which is precisely what these premise gates exist to prevent, just inverted.

Widened at the helper rather than at the call site, since its other two callers want the same
thing: they need `nt`'s *types*, and DIA-backed symbols answer those as well as `pdb` does. The
doc comment now says so.

Re-verified on the ARM64 bench: clippy clean, dump tier **61 passed / 0 failed / 12 ignored**.

PR #160, opened 2026-08-18
Stop paying for instructions the client throws away

glslang:
Two of `FOLLOWUPS.md` item 24's findings, and one correction to the item itself.

## The instructions overran what a client reads

3,147 characters sent, **2,048 read** — so a third of the text was charged for on every connection
and discarded. What fell off the end was the `debug_batch` paragraph: the one instruction there
that stops a mutation being left half-applied. A model was being told about transactional batches
in bytes it never received.

Now **1,996 characters** with that guidance inside the budget, and kept ASCII so the character count
and the byte count cannot diverge — an em dash is three bytes, and a limit stated in characters is
not one to test with multi-byte punctuation. The protocol tier asserts both counts *and* that the
batch guidance survives, so the next edit cannot quietly push it back over.

## `session_id` had three wordings, and five tools had none

`heap_list`, `heap_allocations`, `heap_chunk`, `heap_census` and `heap_diagnostics` documented their
`session_id` not at all — a caller reading those had nothing telling them what to pass. One wording
everywhere now. `server_log`'s keeps its own, because there the field filters records rather than
routing a call.

Together: **67,613 B → 66,320 B** of model-visible surface.

## The item's own number was wrong, instructively

Item 24 claimed **9,514 B** for `session_id` across 43 sites. The model-visible total was
**4,695 B across 32** — the count had included the copies inside `outputSchema`, which the same
document establishes, two sections earlier and from measurement, that no model reads. So the
achievable saving was never ~9 KB, and this change is −1,293 B rather than the ~14% I would have
claimed from the bullet.

A number not measured in the channel it is claimed for is not a finding. Both the item and
`docs/token-budget.md` now say so.

## What the measurement does support

`docs/token-budget.md` gains finding 8: **five tools are a third of the model-visible surface**, and
it is their *input* schemas rather than their prose.

| tool | model-visible | of which inputSchema |
|---|---:|---:|
| `debug_batch` | 9,728 | 7,962 |
| `walk_memory` | 4,058 | 2,695 |
| `crash_triage` | 2,894 | 1,290 |
| `reachable_from_dispatch` | 2,610 | 1,990 |
| `server_log` | 2,599 | 1,608 |
| **total** | **21,889 (33%)** | |

Median tool: 882 B. `debug_batch` alone is 15% of everything a model is handed before it asks
anything, and 7,962 B of that is the `StepAction`/`Check` vocabulary its schema pulls out of
`src/batch.rs`. That is a different kind of problem from the duplication above it — one tool
honestly describing a rich argument — and the levers are design choices (a smaller step vocabulary,
a `$ref` the client resolves, a surface that does not offer every tool to every caller) rather than
tidying.

## Testing

ARM64 bench: clippy clean, `cargo test --test mcp_smoke` **61 passed**,
`WINDBG_MCP_SMOKE_DUMP=1 … --test-threads=1` **61 passed / 0 failed / 12 ignored**. Goldens
re-recorded; `docs/token-budget.md` re-stated from this run with the `2025-06-18` payload
re-measured. `markdownlint-cli2`: 0 issues in 20 files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,7 +125,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **The server instructions now fit what the client reads.** They were 3,147 characters against a
   measured 2,048-character limit, so 1,099 were paid for on every connection and discarded — and
   what fell off the end was the `debug_batch` paragraph, the one instruction there that stops a
-  mutation being left half-applied. Rewritten to 1,996 characters with that guidance inside the
+  mutation being left half-applied. Rewritten to 1,990 characters with that guidance inside the
   budget, kept ASCII so the character and byte counts cannot diverge, and asserted in the protocol
   tier so it cannot grow back unnoticed.
 
@@ -146,7 +146,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   which the same document establishes no model reads. The real model-visible figure was 4,695 B.
   Both the item and [`docs/token-budget.md`](./docs/token-budget.md) are corrected, and the doc
   gains the finding that measurement actually supports — five tools carry a third of the surface,
-  in their *input* schemas, `debug_batch` alone being 15% of it.
+  in their *input* schemas, `debug_batch` alone being 15% of it. The `session_id` wording also gained
+  back what an earlier draft of it dropped: a handle is refused when the target was **replaced** by a
+  raw command, not only when the session closed, and the README explains the asymmetry that makes
+  that guarantee worth having — a retired session refuses its handle while still answering a call
+  that names none.
 
 - **`modules` reports which PDB the engine has for a module** — `guid`, `age`, and the `key` those
   two make, which is the middle element of a symbol server's `<pdb>/<key>/<pdb>` path. It completes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **The server instructions now fit what the client reads.** They were 3,147 characters against a
+  measured 2,048-character limit, so 1,099 were paid for on every connection and discarded — and
+  what fell off the end was the `debug_batch` paragraph, the one instruction there that stops a
+  mutation being left half-applied. Rewritten to 1,996 characters with that guidance inside the
+  budget, kept ASCII so the character and byte counts cannot diverge, and asserted in the protocol
+  tier so it cannot grow back unnoticed.
+
+- **One wording for `session_id`, on every tool that takes one.** It was documented three different
+  ways across 26 tools and not at all on five (`heap_list`, `heap_allocations`, `heap_chunk`,
+  `heap_census`, `heap_diagnostics`), where a caller had no way to know what to pass.
+  `server_log`'s keeps its own wording, because there the field filters records rather than routing
+  a call — a different question that deserves a different sentence.
+
+  The wording also now says what the field *does*. All three originals described only the staleness
+  guard — "pass it to refuse the call if the target has been replaced" — which is the consequence,
+  not the behaviour: the field is a **router**, and omitting it sends the call to the current
+  session, which with several open is not necessarily the one you were last using. That makes this
+  half close to byte-neutral, and the instructions are where the surface actually moves: 67,613 B to
+  **67,076 B** model-visible, plus 1,167 B of instruction text that was being sent and discarded. That is smaller
+  than [`FOLLOWUPS.md`](./FOLLOWUPS.md) item 24 predicted for the `session_id` half, and the reason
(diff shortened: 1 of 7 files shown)


glslang:
Rebased onto `main` (the duplicate handoff commit dropped out), and the review finding is fixed —
it caught a regression I introduced.

**Codex P2 — describe `session_id` as a router.** Right, and worth spelling out: all three original
wordings described only the staleness guard ("pass it to refuse the call if the target has been
replaced"), which is the *consequence*, not the behaviour. `resolve()` routes — omitted, the call
goes to `registry.current()`, the newest session still open; given a handle, to that session, and
the refusal happens only when that session has closed. By unifying on the most common of the three,
I would have propagated the incomplete description to all 31 sites, including the five heap tools
that previously had none — telling a caller that omission is safe when several sessions are open,
which is exactly when it is not.

The new wording leads with routing and keeps the staleness clause second.

**That changes this PR's arithmetic, and the doc now says so.** A correct description of two
behaviours is not shorter than an incomplete description of one, so the `session_id` half is close
to byte-neutral by construction: the model-visible surface goes **67,613 → 67,076 B** (−537), not
the −1,293 I first reported. The instructions are the real movement — 1,167 B that was being sent
and discarded — and the honest read of the whole PR is that item 24's `session_id` bullet was worth
far less than it claimed in both directions: the 9,514 B was miscounted, and the achievable saving
was a wording that turned out to be wrong.

`docs/token-budget.md` finding 3, `FOLLOWUPS.md` item 24 and the CHANGELOG all carry that
correction rather than the flattering version.

Re-verified on the ARM64 bench against the rebased branch: clippy clean, goldens re-recorded,
`WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke -- --test-threads=1` **61 passed / 0 failed /
12 ignored**, `2025-06-18` payload re-measured. `markdownlint-cli2`: 0 issues in 20 files.

glslang:
Both findings fixed, and the second one took the wording to a third draft — worth saying why.

**Codex — document refusal after the target is *replaced*.** Correct, and it is a guarantee I
dropped. `SessionState::accepts_handle` excludes `Retired`, while `accepts_default` includes it, so a
raw `execute` that replaces the target (`.opendump`, `.attach`, `.detach`) leaves a session that
**refuses its handle but still answers a call naming nothing**. That asymmetry is exactly where
passing the id earns its keep, and my "refused if it has closed" described the narrower case.

The first attempt at fixing it spelled the asymmetry out per field and cost **+4,083 B** — worse
than the incomplete wording it replaced, and 31 copies of a paragraph that is true once. It now
reads "refused if its target was replaced or closed" on the field, and the asymmetry itself is in
the README's session section, where it costs nothing per tool. Both misses were the same kind:
describing a consequence instead of the behaviour.

**Codex — the finding-8 numbers were stale against the golden.** Also correct: I wrote them before
the `session_id` rewording moved them. Regenerated from the current recording — 9,746 / 4,076 /
2,912 / 2,628 / 2,599, **21,961 B (33%)**, median **900 B** — in both `docs/token-budget.md` and
`FOLLOWUPS.md`. Exactly the figures you quoted.

Net effect unchanged: 67,613 → **67,076 B**, with the `session_id` half byte-neutral by construction
and the instructions doing the work.

Re-verified on the ARM64 bench: clippy clean, goldens re-recorded, dump tier **61 passed / 0 failed
/ 12 ignored**, `2025-06-18` payload re-measured. `markdownlint-cli2`: 0 issues.

glslang:
Both fixed — and both were the same failure mode as the `session_id` one: compressing prose until a
qualifier that was doing real work fell out.

**Codex — keep the interrupted result on the original call.** Right. My shortened text said
`interrupt` "returns what it had reached and frees the session", which describes neither end
correctly: the interrupt returns as soon as the break is raised, the partial output comes back on
the *original* outstanding call, and the engine can still be busy until that call unwinds. A model
following the old wording would read the wrong response and start follow-up work against a session
that is not free yet. Now: "the interrupt returns at once, and the partial result comes back on the
original call."

**Codex — preserve the `run_to_address` target restriction.** Also right, and worse in context: the
sentence sits in the driver-IOCTL paragraph, which is exactly the workflow someone runs on
`attach_kernel_local` — where the tool cannot work, because a local kernel cannot set code
breakpoints. Now: "which needs a KDNET/VM kernel target, not a local one."

Both had to fit the 2,048-character budget, so I paid for them out of detail each tool already
carries in its own description — `reachable_from_dispatch`'s REACHABLE/NOT REACHABLE asymmetry most
of all, which a model reads when it looks at that tool rather than needing up front. Net: **1,990
characters**, 58 to spare, with two qualifiers restored and nothing lost that is not documented
elsewhere.

Goldens re-recorded; the numbers in `docs/token-budget.md`, `FOLLOWUPS.md` and the CHANGELOG follow
this run (model-visible still 67,076 B; instructions 3,163 → 1,990).

Verified on the ARM64 bench: clippy clean, dump tier **61 passed / 0 failed / 12 ignored**.

PR #161, opened 2026-08-18
Say why a listener started through ssh is gone a second later

glslang:
Found setting up the local-model experiment: the listener was started with a `Start-Process` from an
`ssh` command, bound its port, logged `windbg-mcp listening on http://127.0.0.1:8765`, and was gone
by the time anything connected — with nothing in the log to say so, because from the server's point
of view nothing failed.

**Windows OpenSSH terminates the session's process tree when the session ends**, and `Start-Process`
does not escape it. Measured rather than assumed:

```
inside the ssh session: alive=True pid=5540
after it ended:         still listening on 8799 = False
```

## Why it earns a section rather than a line

The explanation that comes to mind first is wrong, and it is the one I started writing down: *it
lost its stdin*. Under stdio a closed stdin **is** the disconnect and tears every session down;
under `--listen` there is no such signal — that is the whole reason the lease exists — and
`serve_http` is handed a shutdown future that never completes (`main.rs`). Checking that is what
stopped a plausible-sounding wrong mechanism going into the docs as fact, and it is worth writing
down so the next person is pointed at the right half of the server.

The new section gives the failing shape, the note that it is not stdin, and the two arrangements
that work:

- **one ssh channel that both forwards and runs it** — right for a session you are driving anyway,
  since the listener then lives exactly as long as the tunnel and leaves nothing behind;
- **the service role** — right for anything that should outlive a tunnel, a logout or a reboot.

Docs only; no code changed. `markdownlint-cli2` over CI's globs: 0 issues in 20 files.

### DIFF
--- a/docs/remote-listener.md
+++ b/docs/remote-listener.md
@@ -33,6 +33,73 @@ readable by every process on the machine — including a debuggee you `launch`. 
 the environment of both processes this server creates (engine workers, and the TTD recorder), so a
 recorded or launched target cannot read it and claim the server.
 
+### Do not start it *through* ssh and then hang up
+
+The two blocks above assume you are on the Windows machine for the first one. If instead you start
+the listener from an `ssh` command and let that command finish, it dies with the session:
+
+```console
+# on the client machine — starts, binds, logs "listening on ...", and is gone seconds later
+ssh windbg-vm 'powershell -Command "Start-Process windbg-mcp.exe -ArgumentList \"--listen\",\"127.0.0.1:8765\" -WindowStyle Hidden"'
+```
+
+**Windows OpenSSH terminates the session's process tree when the session ends**, and `Start-Process`
+does not escape it. The symptom is confusing because the log looks like a healthy start: the bind
+succeeds, `windbg-mcp listening on http://127.0.0.1:8765` is written, and nothing reports a failure
+— the process is simply not there afterwards, and the port is not bound. Measured: alive at
+`HasExited=False` three seconds in with the ssh session still open, and `Get-NetTCPConnection` on
+the port empty once it closed.
+
+**It is not stdin.** That is the intuitive explanation and it is wrong: the listener role never
+reads stdin. Under stdio, a closed stdin *is* the disconnect and tears every session down; under
+`--listen` there is no such signal, which is what the lease below exists to replace, and
+`serve_http` is handed a shutdown future that never completes (`main.rs`). Worth knowing, because
+"it lost its stdin" sends you looking at the wrong half of the server.
+
+Two ways to keep it up:
+
+**One ssh channel that both forwards and runs it.** Good for a session you are driving anyway — the
+listener lives exactly as long as the tunnel, which is usually what you want, and there is nothing
+left running when you are done:
+
+```console
+ssh -L 8765:127.0.0.1:8765 windbg-vm 'windbg-mcp.exe --listen 127.0.0.1:8765'
+```
+
+The token comes from the environment `setx` put it in, which an ssh session inherits.
+
+**Know what that protects and what it does not.** This server strips `WINDBG_MCP_LISTEN_TOKEN` from
+every child it creates (`engine::spawn_worker`, `ttd::record_launch`), so a debuggee you `launch` or
+a target you `record_trace` does not *inherit* it — which is the accident worth preventing, and the
+common one. It is not a defence against a target that goes looking: `setx` writes to
+`HKCU\Environment`, and a process running as the same user can read that key, a file under that
+user's profile, or the listener's own memory. Nothing the listener can read is hidden from a
+program running as the listener's user.
+
+So the account is the boundary, and **this server cannot put an untrusted target on the other side
+of it**. `launch` creates the process through DbgEng and `record_trace` spawns TTD with `-launch`;
+both run the target as the listener's own account, and there is no alternate-account path in this
+server today. An untrusted binary passed to either can therefore read `HKCU\Environment`, recover
+the token, and claim the listener — which is `execute`, `debug_batch` and `launch` on every session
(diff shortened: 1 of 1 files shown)


glslang:
Both findings were the same mistake and it was a real one, so thank you for the P1.

**The keep-alive recipe read the token out of a file under the user's profile.** That defeats the
protection the section above it describes: this server strips `WINDBG_MCP_LISTEN_TOKEN` from every
child it creates (`engine::spawn_worker`, `ttd::record_launch`) precisely so a `launch`ed debuggee
or a `record_trace` target cannot read it — and a file readable by the user is readable by that
target directly. Handing a debuggee the bearer token lets it authenticate to the listener and take
over the session debugging it, which is the one outcome the whole loopback-plus-token arrangement
exists to prevent.

It was also, as the second comment says, a path nothing else in the repo creates — I invented it for
a local experiment and then wrote it down as if it were the documented setup. Two failures for the
price of one: an insecure recipe *and* one that would not have worked for anyone following the page
in order.

The recipe now just runs the exe, and the ssh session inherits the variable `setx` already put in
the environment — which is what the page established twenty lines earlier. The note says why a file
is not an equivalent substitute, since the reasoning is not obvious and the failure is silent.

`markdownlint-cli2`: 0 issues. Docs only.

glslang:
Both fixed, and the P1 is a correction to my correction — worth being explicit about, because the
first fix looked right and was not.

**Codex — `setx` is not a hiding place either.** Correct. I replaced a token file with the
environment on the grounds that this server strips `WINDBG_MCP_LISTEN_TOKEN` from every child it
creates — but `setx` writes to `HKCU\Environment`, and a process running as that user can read the
key directly, no inheritance required. So the second recipe preserved the risk of the first.

Rather than invent a third hiding place that does not exist, the section now says what is actually
true: **nothing the listener can read is hidden from a program running as the listener's user.** The
environment strip stops the *accident* — a debuggee inheriting the variable, which is the common
case and worth preventing — and it is not a boundary. The boundary is the account: if the targets on
this host are untrusted, and on a machine you debug exploits with they are, the debuggee must not
run as the account the listener runs as. Reaching the token means `execute`, `debug_batch` and
`launch` on every session the listener holds, so the honest framing matters more than a tidy recipe.

Note this applies to the page's existing `setx` instruction too, not only to the block I added — the
new text sits under both.

**Codex — qualify the immediate handover.** Right: `Lease::released` only records `farewell`, and
`try_give_up` keeps the holder while `in_flight != 0` or `Sessions::busy()`. So a `409` just after a
`DELETE` is the departing client's work draining, not a failure, and the paragraph now says so
rather than promising an immediate handover the code does not make.

`markdownlint-cli2`: 0 issues. Docs only.

glslang:
Rebased onto `main`, and the P1 is fixed — it was right, and it is the third time this section has
been corrected in the same direction: I keep proposing a mitigation and the reviewer keeps pointing
out that the mitigation is not available.

**Codex — `launch` and `record_trace` cannot put a target on the other side of the account
boundary.** Verified in the tree rather than taken on trust: `launch` goes through DbgEng's
`CreateProcessWide` (`worker.rs`), `record_trace` spawns TTD with `-launch` (`ttd.rs`), and both run
the target as the listener's own account. There is no alternate-account path. So "run the debuggee
as a different user" was advice a reader could not act on with this server's own tools.

The section now names the limitation instead of the workaround, and says what remains usable:

- **do not `launch` or `record_trace` an untrusted binary** on a host whose listener token matters —
  record it elsewhere and bring the trace over;
- opening a dump or trace executes nothing, so it is unaffected;
- `attach_process` is fine when the target already runs as a different user, because this server did
  not create the process;
- a live kernel target is a different machine entirely, which is what this endpoint is mostly for.

Worth stating plainly: across three rounds this went token-file → `setx` → "use another account" →
"that account cannot be arranged for the two tools that create processes." The first two were
mitigations that did not mitigate. The value of the section now is that it tells you which tools are
safe to point at something untrusted, which is the question a reader actually has.

`markdownlint-cli2`: 0 issues. Docs only.

glslang:
Fixed. **Codex — `end_session` ends one session, not all of them.** Correct, and the advice was
ambiguous in the worst direction: "end the session first" reads as a single action, while a client
holding two live kernels would release one and leave the other owned for the whole grace — which is
precisely the situation the paragraph is warning about.

It now says `end_session` releases **one** target, the one you name or the current one, so a client
with several has to call it for each, and points at `session_status` for the list.

Third finding on the same paragraph, and the pattern is worth noting: every one has been a case of
advice that reads as complete and is not — a mitigation that does not mitigate, an account boundary
that cannot be arranged, and now a cleanup that only cleans up part of it.

`markdownlint-cli2`: 0 issues. Docs only.

glslang:
Fixed, and both halves were right.

**`attach_process` to a different account is not a boundary if the target is privileged.** A process
running as SYSTEM or as an administrator can read the listener's process memory or the user's
registry hive whatever account it nominally holds, so the qualifier had to be
**less-privileged**, not merely different. My bullet would have read as a green light for attaching
to an untrusted service.

**And `attach_kernel_local` makes "a live kernel is another machine" false.** That tool debugs the
host the listener runs on, so an untrusted driver there is inside the boundary like anything else
local. The bullet now names `attach_kernel` over KDNET or serial as the remote case and says
`attach_kernel_local` is not it.

This is the fourth correction to this one paragraph, and each has been the same shape: a statement
that is true of the common case, read as a rule. Token file → `setx` → "use another account" →
"that account cannot be arranged for the tools that spawn" → "and 'different account' is not the
same as 'less privileged', and not every kernel is remote." Security advice that is nearly right is
worse than none, so I would rather it took four rounds here than one in someone's setup.

`markdownlint-cli2`: 0 issues. Docs only.

Issue #163, opened 2026-08-18
Flaky: the rollback's final progress notification races the end_session response


glslang:
## What fails

`ending_a_session_stops_a_running_batch_and_rolls_it_back`, intermittently, in the debugger tier:

```
thread 'ending_a_session_stops_a_running_batch_and_rolls_it_back' panicked at tests/mcp_smoke.rs:4955:
the rollback finished and the client was never told: ["a transaction is in flight; rolling it back (up to 2m03s)"]
```

The client sees the *promise* milestone and never the *retraction*. Seen four times across a working
session; three consecutive full-tier runs in between were green, so it is roughly one in five.

## Why

`Waiting::unwound` in `src/engine.rs` names the mechanism, for a neighbouring problem:

> The worker's two `RollingBack` messages are emitted by **different threads** — the promise by its
> request reader, the retraction by its engine thread — so a teardown landing exactly as a batch
> exits can put them on the wire in either order.

That comment is about *reordering*, which is defended against. The flake is the other consequence of
the same fact: the retraction can arrive **after the `end_session` response has been answered**. A
call's progress sink lives beside that call's waiter, so once the teardown completes there is
nothing left to relay a late milestone to, and it is dropped.

So the ordering is: teardown finishes → response written → retraction arrives → no sink → the client
is never told, exactly as the assertion says.

## Is it a product bug?

Mildly. The tool *result* carries the truth — a caller reading the answer knows the rollback
completed — so what is lost is a progress notification, not a fact. But a client watching progress
on a long unwind sees it stop at "rolling it back (up to 2m03s)" with no closing message, which is
the one moment that milestone exists for.

## Options

1. **Answer after the retraction.** Make the teardown wait for the engine thread's `RollingBack`
   before writing the response. Deterministic, and couples the response to a message that may never
   arrive — a worker killed mid-unwind would then hold the answer until a timeout.
2. **Keep the sink alive briefly past the response** so a late milestone still finds a relay. Small,
   and makes "notifications may arrive after the result" an explicit property rather than an
   accident.
3. **Weaken the test** to take the result text as the source of truth for completion, keeping the
   progress assertion for the promise only. Cheapest, and loses the assertion that caught the
   original defect — the retraction rendering as "a transaction still in flight, up to 0.0s".

I would take 2: it fixes what a watching client actually experiences without making the answer wait
on a message that is not guaranteed.

## Not caused by

Not related to `#162`'s idle release or the coordinate work — first seen on `main` before either,
[...]

Issue #162, opened 2026-08-18
Single-tenancy and the lease are keyed on Mcp-Session-Id, which 2026-07-28 removed


glslang:
## What happens

Against `--listen`, at protocol revision **`2026-07-28`** — the newest this server advertises, and
the one Claude Code negotiates — **two clients are both served**, and neither ever holds a lease.

```
Two independent clients at 2026-07-28 (sessions removed from the spec):
  client A tools/call -> HTTP 200  No debug session is open. Start one with open_dump / ...
  client B tools/call -> HTTP 200  No debug session is open. Start one with open_dump / ...

(a 409 for B is what the single-tenancy gate looks like on 2025-06-18)
```

On `2025-06-18` the same probe gets `409` for the second client, as designed.

## Why

Per **SEP-2567**, sessions are removed from `2026-07-28`. rmcp implements this — from
`StreamableHttpServerConfig::legacy_session_mode`:

> Only applies to legacy protocol versions (`< 2026-07-28`). Per SEP-2567, sessions are removed from
> the `2026-07-28` version, so requests negotiating that version are always served statelessly
> regardless of this setting.

So no `Mcp-Session-Id` is minted or accepted. `listen.rs` reads that header to identify a client
(`SESSION_HEADER`, `Lease::admit`, `Lease::settle`), so on the newest revision every request arrives
as `session: None`:

- `admit(None)` matches the *new client* arm and is served;
- `settle(minted: None, requested: None)` falls through the match and **installs no holder**.

Tenancy is therefore never taken, and — because a deadline is only set on the holder paths — the
lease never counts down either.

## Why it matters

Both properties this was built for are silently absent on the revision most clients will pick:

- **Single tenancy.** `docs/remote-listener.md` says why it exists: "session handles are minted
  globally, the four-session cap is shared, and `end_session` ends whatever it is handed — so two
  clients would silently share, and one could end a target the other was using." That is now the
  default behaviour, not the refused one.
- **The lease.** "A client that has stopped talking is indistinguishable from one that is thinking"
  is solved by the lease releasing that client's sessions. With no holder there is nothing to
  expire, so a client that vanishes leaves its sessions — and their engine workers — held until
  `end_session` or process exit. For a live kernel that means the target stays owned indefinitely,
  which is the exact property phase 1 was written to rebuild.

The rest of the transport is fine statelessly: `tools/list` and `tools/call` both work at
[...]

PR #164, opened 2026-08-18
Release a session nobody is using, since no disconnect is coming

glslang:
First slice of [#162](https://github.com/glslang/windbg-mcp/issues/162), and deliberately the one
that needs no design decision.

## The problem

`2026-07-28` removed sessions from MCP (SEP-2567). The lease identifies a client by
`Mcp-Session-Id`, so on the revision most clients now negotiate there is no holder to install, no
lease to expire, and **no moment at which a client can be said to have gone away**. A client that
vanishes leaves its targets held until the process exits — for a live kernel, a machine owned by
nobody, which is the property phase 1 was built to rebuild.

Measured, not inferred:

```
Two independent clients at 2026-07-28 (sessions removed from the spec):
  client A tools/call -> HTTP 200
  client B tools/call -> HTTP 200      (2025-06-18 gives B a 409)
```

## What this adds

The lease was doing two jobs and only one needs a client identity. "Has *this client* gone" cannot
be answered without one. "Is anyone still using this target" can, and that is this: a session unused
for **30 minutes** is released, through the same orderly `EndSession` an explicit `end_session`
uses, because a live kernel that is merely killed is left halted.

| | |
| --- | --- |
| Default | 30 minutes since the last call naming that session |
| Override | `WINDBG_MCP_SESSION_IDLE_SECS`, whole seconds; `0` disables |
| Floor | must exceed the longest a single call can run, or the server refuses to start |

## Three things it must not get wrong

**A session with a call outstanding is never idle**, however old the call. Not defensive coding: an
`attach_kernel` parked in `WaitForEvent(INFINITE)` submits one call and then waits for its target to
dial in, possibly for hours, and it is exactly the session that must survive. Reading the clock
alone would release the one session whose entire job is to wait. `Session::busy()` already knew
this; the idle check asks it first, and a test backdates a session an hour to prove an outstanding
waiter still saves it.

**Nor is an opener that has not handed its handle back** — nobody can have used a session whose id
its caller has not been told.

**The timeout has the same floor the lease refuses to start below.** A single call can keep a
session quiet for its whole budget, so anything shorter releases a session underneath its own
caller. Misconfiguration is a startup message, not a target that disappears mid-analysis.

## Why 30 minutes rather than the lease's 390 seconds

They measure different things. A lease is renewed by *any* request, so a working client renews it
constantly. This is per session, and twenty minutes of reading a stack before the next question is
ordinary work, not abandonment. A backstop, not a scheduler.

## What it does not do

It does not make two clients safe from each other — that needs the client identity and the
namespaces built on it, slices 2 and 3 in #162. It stops the leak that has no workaround today, on
every transport rather than only the one with sessions in it.

## Testing

ARM64 bench: clippy clean, `cargo test --bins` **457 passed** (six new — three for the idle
decision, three for the configuration floor and the `0` case).

`WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke -- --test-threads=1` hit
[#163](https://github.com/glslang/windbg-mcp/issues/163) once — a pre-existing flake in the batch
rollback's *progress notification*, unrelated to this change and first seen on `main` before it.
That issue now carries the diagnosis: the worker's two `RollingBack` messages come from different
threads, and the retraction can arrive after the `end_session` response, by which point the call's
progress sink is gone.

## Summary by CodeRabbit

* **New Features**
  * Sessions are now automatically released after 30 minutes of inactivity by default.
  * The idle-release period can be configured with `WINDBG_MCP_SESSION_IDLE_SECS`; setting it to `0` disables the feature.
  * Active calls and unfinished operations are protected from automatic release.

* **Documentation**
  * Added guidance on idle-session release, configuration, validation, and behaviour when session identifiers are unavailable.
  * Documented the new unreleased changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **A session nobody is using is released, whatever the transport did.** The lease answers "has this
+  *client* gone away" by identifying it from `Mcp-Session-Id` — and `2026-07-28` removed the protocol-level
+  MCP session (SEP-2567), so on the revision most clients now negotiate no holder is ever installed and
+  no lease ever expires. A client that vanished left its targets held until the process exited,
+  which for a live kernel is a machine owned by nobody
+  ([#162](https://github.com/glslang/windbg-mcp/issues/162)).
+
+  The listener now also releases any session that has gone unused for **30 minutes** (
+  `WINDBG_MCP_SESSION_IDLE_SECS`, `0` disables), through the same orderly `EndSession` an explicit
+  `end_session` uses. It needs no notion of a client, which is the point: it is the half of the
+  lease's job that survives a protocol with no sessions in it.
+
+  - **Far longer than the lease grace, deliberately.** A lease is renewed by any request; this is
+    per session, and twenty minutes of reading a stack before the next question is ordinary work.
+  - **A session with a call outstanding is never idle**, however old the call — which is exactly an
+    `attach_kernel` parked in `WaitForEvent(INFINITE)` waiting for its target to dial in, the one
+    session that must never be released. Nor is an opener that has not yet handed back its handle.
+  - The floor is the same one the lease refuses to start below: longer than a single call can run,
+    or a session is released underneath its own caller.
+
+  This is the first of the slices in #162. It does not make two clients safe from each other — that
+  needs a client identity, which a stateless protocol only gets from authentication — but it stops
+  the leak that has no workaround today.
+
 - **The server instructions now fit what the client reads.** They were 3,147 characters against a
   measured 2,048-character limit, so 1,099 were paid for on every connection and discarded — and
   what fell off the end was the `debug_batch` paragraph, the one instruction there that stops a
--- a/docs/remote-listener.md
+++ b/docs/remote-listener.md
@@ -151,6 +151,39 @@ client restart costs a KDNET attach, and a KDNET attach costs a reboot of the ta
 how many sessions were inherited when that happens — or that nothing was open, since a client can
 also arrive to find the previous one had let go of an empty server.
 
+## A session nobody is using is released
+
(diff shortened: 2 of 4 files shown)


glslang:
@coderabbitai review

glslang:
All four fixed. The first is a real race in my own code, and my comment claimed the opposite.

**Codex — claim idle sessions atomically.** Correct. `release_idle` checked `idle_for` and then
called `end`, and `call_as` takes the registry lock to register a waiter — so a call arriving in the
gap would be routed to a session the sweep was about to tear down, losing an operation a caller had
just started. Worse, my doc comment asserted the check was "repeated under the registry lock inside
`release`", which it is not; I wrote the defence I meant to have rather than the one I built.

It now uses the idiom already in the file: claim under the registry lock by transitioning the
session to `Closed` (`claim_overage_victim` does exactly this), then release outside it. Since
`resolve` refuses a `Closed` session, a call is on one side of the claim or the other — the gap is
unreachable rather than narrow.

**Codex — the env-var tests race each other.** Also correct, and the repo already had the answer:
`kdconn::env_entries` is split out with a comment saying why — `set_var` is `unsafe` in edition 2024
and mutates state the whole test binary shares. The configuration is now `idle_release_from(value,
call_timeout)`, pure and handed its input, with `idle_release` reading the environment and
delegating. No `unsafe` in the tests, no shared state, and a fourth case worth having falls out
free: a value that is not seconds is refused by name rather than silently taking the default.

**CodeRabbit — protocol-level session vs this server's sessions.** Right, and the ambiguity is
genuinely confusing in a document that uses "session" for both. SEP-2567 removes the *protocol-level*
MCP session; this server's `Session` handles and `end_session` are untouched. Both the doc and the
CHANGELOG now say so.

**CodeRabbit — the SEP link pointed at the spec.** Fixed, now
<https://modelcontextprotocol.io/seps/2567-sessionless-mcp>.

ARM64 bench: clippy clean, `cargo test --bins` **458 passed**.

PR #166, opened 2026-08-18
Give each client its own sessions, since the protocol no longer names one

glslang:
Slices 2 and 3 of [#162](https://github.com/glslang/windbg-mcp/issues/162).

The registry is one map for the whole server — handles minted from it, the cap shared,
`end_session` ending whatever it is handed — and the listener stood in for a boundary by serving one
client at a time. `2026-07-28` removed the protocol-level MCP session (SEP-2567), so the identifier
that gate was keyed on is gone, and on the revision most clients now negotiate **two clients are
both served and neither can be told from the other**.

## Identity

**In a stateless protocol the only sound identity is authentication.** No session id by design; no
connection to key on, since requests arrive on whatever socket a client's pool hands them; and
`clientInfo` is not retained between requests. The credential is what is left.

```console
setx WINDBG_MCP_LISTEN_TOKEN        "<a long random string>"   # names the client `local`
setx WINDBG_MCP_LISTEN_TOKEN_CI     "<another>"                # names `ci`
```

A name only the holder of a token can present is a boundary. A name a client picks for itself — a
header, `clientInfo` — would be a label, which is why neither was an option.

## What ownership changes

| | |
| --- | --- |
| Routing | a handle routes only for its owner; omitting one finds *that* client's newest session |
| Refusal | another client's handle comes back **unknown**, not "someone else's" |
| Listing | `session_status` shows the caller's sessions and no others |
| Capacity | the four-session cap is per client |
| Reclamation | only ever takes the **same** client's idle session |

The refusal wording is deliberate: confirming that a handle exists but belongs to someone else would
leak that a second client is here and what it holds, and there is nothing the caller could do with
the distinction.

Reclamation is the sharpest of these. Ending another client's target to make room for mine is
exactly the harm a shared registry did — and a shared *cap* is the same harm spelled as denial,
which is why the limit moved too.

## Two details worth keeping

**Stdio calls run as `local`** rather than as a special case, so one set of registry rules serves
both transports instead of one rule and an exception.

**One token configured for two names is refused at startup.** The winner would be a `HashMap`
ordering detail, so the same configuration could put a client's sessions under a different name from
one run to the next.

The identity reaches the engine the way a progress sink does — a task-local set around the whole MCP
call — because threading it through would put a parameter on every tool signature naming something
none of them uses.

## Scope

This separates clients; it does not rank them. Everyone who can authenticate still has the whole
tool surface, `execute` and `launch` included.

The tenancy gate is now **redundant rather than wrong**, so retiring it — with the 409, the lease
and the machinery that exists only because the registry was global — is the next change rather than
this one.

## Testing

ARM64 bench: clippy clean, `cargo test --bins` **467 passed**, including eight new. The three that
carry the claim:

- `a_handle_is_not_usable_by_another_client` — routes for its owner by handle and by default, and
  for anyone else is *unknown* rather than refused;
- `session_status_shows_only_the_callers_own`;
- `one_clients_sessions_do_not_fill_anothers_quota` — a client at its own limit is refused while a
  quiet one still has room.

Plus five for the credential parsing, including the one that caught a real bug while I wrote it:
`WINDBG_MCP_LISTEN_TOKEN_FILE` shares the prefix and would otherwise have configured a client called
`file` whose token is a *path* — with the real token silently absent.

`WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke -- --test-threads=1`: **61 passed**.

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple authenticated clients using named bearer tokens.
  * Isolated sessions, routing, status, capacity limits, reclamation and logs per client.
  * Added optional token-file support and retained local identity for standard input/output connections.
  * Preserved the full tool set for every authenticated client.

* **Bug Fixes**
  * Prevented access to sessions and activity owned by other clients.
  * Rejected duplicate token assignments at startup.
  * Removed credential variables from spawned processes.

* **Documentation**
  * Updated remote listener guidance for multi-client token ownership and session handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     still report another client's activity.
   - **A lease expiry releases the sessions of the client whose lease ran out**, not every session.
     Before ownership those were the same set, because the gate served one client at a time.
+  - The rule for identity is **ambient inside a call, by name outside one**. A tool body reads the
+    caller from the task-local, which is who is asking; the listener's own diagnostics run after
+    that scope has closed and take the client as a parameter, so the one that reports an adoption on
+    reconnect counts the reconnecting client's sessions rather than `local`'s.
   - **The tenancy itself is per client**, so two clients no longer queue behind each other. The gate
     serialised the server when the registry was global and one client could end another's targets;
     keeping it shared would have meant one client's four-minute pool walk making every other client
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     still report another client's activity.
   - **A lease expiry releases the sessions of the client whose lease ran out**, not every session.
     Before ownership those were the same set, because the gate served one client at a time.
+  - **An `Mcp-Session-Id` another client holds is reported unknown**, and the check runs before the
+    caller's own tenancy is consulted. The MCP service keeps one session table for the server, so on
+    a legacy revision the id was the only thing between a client and another's MCP session — and a
+    `DELETE` on it closed that session while the lease that minted it still held the id, leaving its
+    owner failing every request and refused its own re-`initialize` for a grace period.
   - The rule for identity is **ambient inside a call, by name outside one**. A tool body reads the
     caller from the task-local, which is who is asking; the listener's own diagnostics run after
     that scope has closed and take the client as a parameter, so the one that reports an adoption on
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   so there is no connection either; `clientInfo` is not retained. The credential is what is left,
   and a name only the holder of a token can present is a boundary — where a name a client picks for
   itself would be a label. Configuring one token for two names is refused at startup rather than
-  resolved, because the winner would be a hash-map ordering detail.
+  resolved, because the winner would be a hash-map ordering detail. Both refusals name the
+  *variables* to change and never the token: they are printed to stderr, and under the service to a
+  log file, so quoting the credential would leave a working one there.
 
   This separates clients; it does not rank them. Everyone who can authenticate still has the whole
   tool surface.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,45 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Per-client session namespaces, so two clients on one listener cannot reach each other's
+  targets** ([#162](https://github.com/glslang/windbg-mcp/issues/162), slices 2 and 3). A listener
+  may now hold several bearer tokens — `WINDBG_MCP_LISTEN_TOKEN_<NAME>` names a client, the unnamed
+  variable names `local` — and a session belongs to whichever client opened it.
+
+  - **Routing** is per client: a handle only routes for its owner, and omitting one finds that
+    client's newest session.
+  - **Another client's handle is reported unknown**, not "someone else's". The answer must not
+    confirm a session the caller may not touch, and there is nothing they could do with the
+    distinction.
+  - **`session_status` lists only the caller's**, and the four-session **cap is per client**, so a
+    busy client cannot deny a quiet one. A session is only ever reclaimed to make room for its own
+    client — reclaiming another's is the precise harm a shared registry did.
+  - **`server_log`** shows the caller's sessions' records and the supervisor's own — one ring serves
+    the whole server, so without this a client could read what another's worker printed.
+  - **A lease expiry releases the sessions of the client whose lease ran out**, not every session.
+    Before ownership those were the same set, because the gate served one client at a time.
+  - **The tenancy itself is per client**, so two clients no longer queue behind each other. The gate
+    serialised the server when the registry was global and one client could end another's targets;
+    keeping it shared would have meant one client's four-minute pool walk making every other client
+    wait for a boundary the registry now provides properly — namespaces that cannot be used
+    concurrently. Two connections presenting the *same* token still contend, which is one client
+    racing itself.
+  - Under **stdio** everything runs as `local`, so one set of registry rules serves both transports
+    rather than one rule and an exception.
+  - **Every credential variable is stripped from the processes this server creates** — by prefix,
+    so a token added later cannot quietly reach a debuggee — and a configured token *file* shuts the
+    environment out entirely, which is the precedence a LocalSystem service depends on.
+
+  **Why authentication is the identity.** `2026-07-28` removed the protocol-level MCP session, so
+  there is no session id to key on; requests arrive on whatever socket a client's pool hands them,
+  so there is no connection either; `clientInfo` is not retained. The credential is what is left,
+  and a name only the holder of a token can present is a boundary — where a name a client picks for
+  itself would be a label. Configuring one token for two names is refused at startup rather than
+  resolved, because the winner would be a hash-map ordering detail.
+
+  This separates clients; it does not rank them. Everyone who can authenticate still has the whole
+  tool surface.
+
 - **`server_log` — the server's own log, readable from wherever the client is.** The supervisor
   keeps a bounded ring of the most recent records, its own and every engine worker's, and serves
   them as a tool: filterable by session and level, paged with a `since` cursor, answered without
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - **`session_status` lists only the caller's**, and the four-session **cap is per client**, so a
     busy client cannot deny a quiet one. A session is only ever reclaimed to make room for its own
     client — reclaiming another's is the precise harm a shared registry did.
+  - **Closed-session history is per client too**, so a handle still answers after its target is gone
+    however busy the rest of the server has been. A shared bound is a shared fate: one client's
+    churn would age out another's record of a session that failed, and the answer that client then
+    gets for a handle it is still holding is "unknown" — which reads as "never existed".
   - **`server_log`** shows the caller's sessions' records and the supervisor's own — one ring serves
-    the whole server, so without this a client could read what another's worker printed.
+    the whole server, so without this a client could read what another's worker printed. **Its
+    counts are the caller's too**: how full the buffer is, where the cursor is now and what the
+    oldest record is are all over that client's stream, since numbers about records nobody may read
+    still report another client's activity.
   - **A lease expiry releases the sessions of the client whose lease ran out**, not every session.
     Before ownership those were the same set, because the gate served one client at a time.
   - **The tenancy itself is per client**, so two clients no longer queue behind each other. The gate
--- a/docs/remote-listener.md
+++ b/docs/remote-listener.md
@@ -212,6 +212,7 @@ What ownership buys, and it is the whole of it:
 | History | a closed session ages out of `session_status` on its **own** client's churn, not the server's |
 | The log | `server_log` shows the caller's sessions' records, plus the supervisor's own, which name no session — and the buffer counts it reports are over the records that caller can read |
 | Lease expiry | releases the sessions of the client whose lease ran out, and no others |
+| Session ids | an `Mcp-Session-Id` another client holds is reported **unknown**, before this caller's own tenancy is consulted |
 | Contention | **within** a client only — one client's long call does not make another wait |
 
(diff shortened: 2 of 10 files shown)


glslang:
@coderabbitai review

glslang:
All five addressed. Three were security findings and each was real — the critical one is a hazard
this PR widened, which is the worst kind.

**CodeRabbit (critical) — named tokens reached child processes.** Right. `spawn_worker` and
`record_launch` stripped `WINDBG_MCP_LISTEN_TOKEN` by name, and this PR made that a *prefix* of
several credentials, so `…_TOKEN_CI` and `…_TOKEN_FILE` would have gone straight to an engine worker
and any debuggee it launched. Now stripped by prefix, in one helper both call sites share, so a
token added later cannot quietly get through. That is the difference between a strip that enumerates
and one that describes.

**Codex (P1) — token-file precedence lost.** Also right, and I had not noticed I was changing it. A
configured file now shuts the environment out *entirely*, named tokens included — the installer
ACLs that file to SYSTEM and Administrators precisely because the machine environment is readable,
so a variable standing beside it reintroduces what the file exists to avoid.

**Codex (P1) — the lease released across clients.** The sharpest of the three, and only reachable
*because* of this PR: the lease is keyed on an MCP session id, so a stateless client never becomes
the holder while its targets sit in the same registry — one legacy client's timeout would have ended
them. The tenancy now carries its client, and an expiry releases that owner's sessions and no
others.

**Codex (P2) — `server_log` ignored the boundary.** Correct: one ring serves the whole server, so a
client could read another's session ids and worker output, and a foreign `session_id` filtered
straight to it. Records are now narrowed to the caller's sessions, with the supervisor's own — which
name no session — visible to everyone, since they are about this process rather than anyone's
target. A foreign id narrows to a set the caller is not in and comes back empty: the same answer an
id this server never issued gets, for the same reason as `resolve`.

**CodeRabbit (trivial) — module doc length.** Partially taken. The transport rationale that belongs
to operators moved to `docs/remote-listener.md`; what stays in `client.rs` is why *this* design was
forced — no session id, no connection, `clientInfo` not retained — because that is the question a
reader of the code asks, and this repo's own comments answer that kind of question rather than
restating the code.

One self-inflicted note: changing the lease's log line to name the client broke a smoke test that
matched the old wording. The needle is now the stable half of the sentence.

ARM64 bench: clippy clean, `cargo test --bins` **469 passed**, dump tier **61 passed / 0 failed / 12
ignored**.

glslang:
Fixed, and it is a better catch than its P2 badge suggests — the two halves fail in opposite
directions.

**Codex — compare credential names case-insensitively.** Windows environment names are, and
`std::env::var` honours that, so a host configured as `Windbg_Mcp_Listen_Token` worked before this
PR replaced that lookup with a scan. My `strip_prefix` was case-sensitive, which means:

- a listener on such a host **refuses to start**, no longer seeing its own token; and
- worse, `token_file()` still uses `var_os` and *would* resolve a mixed-case
  `…_TOKEN_FILE`, while `strip_credentials` walked straight past it — handing a debuggee the
  credential the strip exists to withhold.

So the same oversight produced a startup regression and a hole in the fix I had just made for the
critical finding. Both scans now fold case, through one `credential_suffix` helper so they cannot
drift apart again, with a test covering `Windbg_Mcp_Listen_Token`, `windbg_mcp_listen_token_ci` and
the mixed-case file variable.

Worth noting what this run of findings has in common: every one has been about the *edges* of the
new credential surface — what else shares the prefix, what precedence it displaces, what inherits
it, and now how it is spelled. Widening a credential from one variable to a family turns out to be
mostly edge.

ARM64 bench: clippy clean, `cargo test --bins` **470 passed**.

glslang:
Both fixed. The first absorbs most of what slice 3b was going to be, because the finding is right
that it cannot sensibly be deferred.

**Codex — lease admission still serialised across clients.** Correct, and it makes this PR's headline
feature unusable: namespaces you cannot use concurrently are not namespaces. One client's pool walk
or `!analyze` would have made every other client wait — for a boundary the registry now provides
properly.

**The tenancy is now per client** rather than per server. The gate still arbitrates *within* a
client, because two connections presenting the same token really are one client racing itself, and
the claim machinery exists for exactly that. Across clients there is nothing left to arbitrate.

Mechanically this is one `Tenancy` per client behind a `TenancyGuard` that derefs to it, so every
rule below reads as it did — the change is entirely in *whose* tenancy is consulted. The sweeper
walks each client in turn. `holder_client`, which I added in the previous round to scope the
release, is gone again: the map key is the client, so carrying it inside the value was a second copy
of the same fact.

Test-side, the ~90 existing lease tests are unchanged: a `For` binding fixes the client once, which
also keeps visible that those tests are about how *a* tenancy behaves rather than about which client
holds it.

**And I verified the new test discriminates**, since a test for "two clients do not contend" passes
trivially if it accidentally builds two leases — which my first attempt did. Against a temporarily
re-globalised `state_of`:

```
test one_clients_tenancy_does_not_block_another ... FAILED
  another client must not wait on a tenancy that no longer bounds anything it can reach
```

**Codex — two credentials normalising to one client.** Also right, and the quieter half of the
duplicate-token problem: nothing looks wrong, both tokens authenticate, and their holders silently
share routing, listing, capacity and teardown rights. `WINDBG_MCP_LISTEN_TOKEN` and `…_TOKEN_LOCAL`
both fold to `local`; `…_CI` and `…__CI` both to `ci`. Refused at startup now, both directions
covered by tests — a boundary two credentials can stand on is not a boundary.

What remains of slice 3b is smaller than it was: whether the gate earns its place *at all* now that
contention is only within a client, which is a question worth asking separately rather than
answering in this PR.

ARM64 bench: clippy clean, `cargo test --bins` **472 passed**, dump tier **61 passed / 0 failed / 12
ignored**.

glslang:
Both fixed, and they are the same finding twice: a per-client boundary with a bound underneath it
that is still counted once for everyone.

**Codex — partition settled-session history by client.** Right, and the consequence is worse than
losing a log line. `session_status` answering **unknown** is the answer that tells a caller to open
again, and for an attach or a launch that means a *second target* — so a quiet client's handle could
be turned into a duplicate kernel attach by a busy client's churn on the other side of the boundary
this PR is about. `trim` now walks the deque client by client, so what ages a client's history out
is that client's own activity.

It is still a bound rather than a hole: the clients are the configured credentials, a set fixed at
startup, so the worst case is `credentials × (CLOSED_HISTORY + MAX_SESSIONS)`. I did **not** make
the live-session cap per client in the same way — `MAX_SESSIONS` is a real resource bound (each
session is a process holding a target), so the four are the machine's, not a policy allowance, and
that one is deliberately shared. What the finding is about is history, which costs a struct.

**Codex — hide foreign records from log pagination metadata.** Also right, and it is the half of the
boundary I built and then reported around. The filter was correct; the numbers beside it were the
*ring's* — `held`, `next_since` and `oldest_seq` all moved with whatever anyone's worker filed. So a
client could poll a quiet log and watch all three climb, which is a report of a neighbour's activity
arrived at without reading one of its records. `describe_log` even renders it in words: "the buffer
holds N record(s) that this filter excluded".

All three are now computed over the caller-visible stream, and `next_since` advances past what that
caller *could have been shown* rather than past what the ring has filed. Two of them got more
accurate in the process — the eviction warning ("records before seq N were evicted since your last
call") used to fire on someone else's records being dropped, and now fires only when the caller has
actually lost something.

**What I did not do, and why:** per-owner sequence *numbering*. The counter is what makes a `since`
cursor exact under eviction, and the ring cannot number per client without knowing which client owns
each session — that fact lives in the registry and arrives here as `Query::visible`, one query at a
time; a query-time ordinal would shift under eviction and break the cursor it exists to serve. So a
client reading two of its own records numbered a hundred apart can still tell that *something* was
filed in between. It learns a count. The records, their sessions and their text stay unreachable,
and it is now written down in `Entry::seq` and in `docs/remote-listener.md` as a limit rather than
left as an omission.

Both tests were checked against the unfixed code before being kept: the history one fails with the
quiet client's sessions evicted, the log one with the buffer looking seven records fuller than its
caller can read.

Verified on the ARM64 bench: clippy clean, 474 unit tests, and the dump-gated smoke tier 61 passed /
0 failed.

glslang:
@coderabbitai review

glslang:
Fixed, and this one is worth more than its P2: the refusal is correct, and it was writing the
credential it refused into the place least able to forget it.

**Codex — redact bearer tokens from collision errors.** Right. The message interpolated the *first*
token, and both places that message goes are durable in a way the misconfiguration is not — stderr
in the foreground, and under the SCM `%ProgramData%\windbg-mcp\service.log`. So a listener that was
misconfigured for one minute leaves a working token in a file that lives for months, readable by
whoever collects logs rather than by whoever was trusted with the credential.

The map behind the check now holds **the variable that configured each client** rather than its
token, so the refusal reads `WINDBG_MCP_LISTEN_TOKEN_CI` and `WINDBG_MCP_LISTEN_TOKEN__CI`. That is
also the half the operator can act on — the token value identified nothing they could go and change.
I did the same for the other direction (one credential under two variables), which named the two
*clients* but not the variables, so it said which boundary had collapsed without saying where.

`a_refusal_never_quotes_the_token_it_refuses` walks both directions with recognisable secrets and
asserts neither appears. Verified against the old message first — it fails with
``(`s3cret-alpha` and `WINDBG_MCP_LISTEN_TOKEN__CI`)``.

Worth noting where this does *not* apply: the connection-string redaction in `kdconn` is a different
mechanism for a different reason (a key that has to travel), and a profile-based attach never puts
one in a message at all. This was the one place a secret was being formatted for a human.

ARM64 bench: clippy clean, 475 unit tests, dump-gated smoke tier 61 passed / 0 failed.

glslang:
@coderabbitai review

glslang:
Fixed. The finding is right, and the second of the two outcomes it names is the worse one: with
`local` holding sessions, the line reports **`local`'s count** to a named client reconnecting — and
the adoption line is a supervisor record, which carries no session id and is therefore readable by
every authenticated client through `server_log`. So the diagnostic reached across the boundary the
rest of the PR builds.

`Sessions::live_count_for(&client)` takes the owner as a parameter. What the fix settles is a rule
worth having written down, since this is the first place it bit:

> **Ambient inside a call, by name outside one.**

A tool body reads the identity from the task-local because that is who is asking, and not threading
it through forty tool signatures is the whole reason it is a task-local. The listener's own
diagnostics run *after* `as_client` has closed — this one reports on a call already handled — so
there the caller is the only thing that still knows, and it has to say so.

I deliberately did not make `snapshot()` take an owner as well. Its one production caller is
`session_status`, inside a tool body, where the ambient answer is correct and explicit passing would
be the parameter the design avoided. The distinction is position, not the method.

**On coverage, plainly:** the new test asserts the count is right when asked by name *and* that
`snapshot()` from the same position is empty — the answer the old call site was getting — so it
pins the mechanism. It does not exercise the HTTP handler itself: an end-to-end guard would need the
listener smoke tier to run a *named* token through open → `DELETE` → reconnect-within-grace and then
read the line back out of `server_log`, which the tier does not do today (it uses the unnamed token
throughout). That is worth having and is a bigger piece than this fix; say the word and I will file
it rather than bolt it on here.

ARM64 bench: clippy clean, 476 unit tests, dump-gated smoke tier 61 passed / 0 failed.

glslang:
@coderabbitai review

glslang:
Fixed, and this is the best finding of the round — it is the one place the boundary was built on top
of something still global rather than beside it.

**Codex — bind MCP session IDs to the authenticated client.** Right. The service keeps one session
table for the whole server, and per-client tenancy did not partition it, so a client presenting
another's `Mcp-Session-Id` reached that client's MCP session; the task-local only decided which
*debug* sessions the tools then saw. The `DELETE` path is the sharp end exactly as described — rmcp
closes the session while the lease that minted the id still holds it, so its owner fails every
request and its own re-`initialize` is refused for a grace period. One authenticated client denying
another is the harm ownership is here to stop.

**The check is in `admit`, not beside it.** That is the one gate every request already passes, and
it runs before the caller's own tenancy is consulted; a check in the handler would be one somebody
can later forget to make. The new `Admission::NotYours` becomes `404 unknown session` — the registry
already reports another client's *debug* handle as unknown rather than refused, for the same reason
(the answer must not confirm a session the caller may not use), and 404 is what the spec gives a
session that has gone, so a client is told to start again, which is the right advice either way.

**Only the holder counts**, and that boundary is deliberate: an id no tenancy holds is already
closed in the service — by the `DELETE` that departed, or by the sweep that expired the lease — so
treating it as owned would refuse a client the resume inside the grace that the arm underneath it
exists to allow. The test covers all three: another client's id is `NotYours`, the holder's own id
still serves, and an id nobody holds still serves.

Verified against the check disabled, where the test fails with
`Serve(Admitted { claim: Some(1), epoch: 0 })` — served *and* holding a claim, which is the shape of
the denial in the finding.

Scope note: this only ever applied to a client on a **legacy** revision, since `2026-07-28` has no
`Mcp-Session-Id` to present. It is worth having anyway — legacy is exactly the configuration where
the lease still matters.

ARM64 bench: clippy clean, 477 unit tests, dump-gated smoke tier 61 passed / 0 failed.

glslang:
@coderabbitai review

PR #167, opened 2026-08-19
Say one client per credential everywhere it was still one client per server

glslang:
Handoff pass after #166. That PR documented the boundary it built; this one fixes the places that
still describe the boundary it replaced — and adds the section a future agent needs before touching
the code.

### The rule worth writing down

`CLAUDE.md` gains a section on serving several clients, whose load-bearing part is:

> **Identity is ambient inside a call and by name outside one.**

`client::current()` reads a task-local that `serve_one` sets around the whole MCP call, which is why
no tool signature carries a caller. Anything running *outside* that scope — the listener's own
diagnostics, a sweep, a shutdown — gets the default `local` rather than an error, so it has to take
the client as a parameter. That is not a hypothetical: it is where #166's last review round found a
log line reporting `local`'s session count to a named client on reconnect.

The section also covers what bites: two tokens on one host are two namespaces, so a session that
"vanished" is usually a request that carried the other token; credentials are built from variables
handed in rather than read from the environment (`set_var` is `unsafe` in edition 2024); they are
stripped from child processes **by prefix**; and neither startup collision may quote a token.

### Claims that were true and are not

- **The `409` body** said `another client holds this server; it serves one at a time`. Tenancy is
  per client now, so that response is only ever one client racing itself — the old wording sends an
  operator looking for a colleague who is not there. It now says
  `this credential already holds a session here; one at a time per client`.
- **`README.md`** pointed at the listener doc for "why only one client is served at a time".
- **The smoke tier's tenancy test** was named `a_second_client_is_refused_…`, where its listener
  holds one token and both connections are therefore the same client. Renamed to what it asserts,
  and its rationale — "two clients would silently share, and one could end a target the other was
  using" — is now history rather than the reason.
- **`docs/smoke-test.md`** now says out loud that every listener assertion runs a single, unnamed
  token, so what they prove they prove for `local` alone.

### Two follow-ups filed rather than fixed

- **28 — the tenancy gate may no longer earn its place.** What is left for it to arbitrate is one
  client racing itself. The item is written with the argument on both sides, because the lease is
  *also* what releases a vanished client's targets, and retiring the gate must not retire the sweep.
- **29 — the listener tier has never run a named client.** Which is how the adoption-count bug got
  as far as review. The item says what a two-token tier would assert, and that all of it is
  dump-tier work.

Verified on the ARM64 bench: clippy clean, 477 unit tests, dump-gated smoke tier 61 passed / 0
failed — including the renamed test and the `409` body assertion, which matches on the stable
substring rather than the sentence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     keeping it shared would have meant one client's four-minute pool walk making every other client
     wait for a boundary the registry now provides properly — namespaces that cannot be used
     concurrently. Two connections presenting the *same* token still contend, which is one client
-    racing itself.
+    racing itself — and the `409` says so, where it used to report a second client that no longer
+    exists as a category.
   - Under **stdio** everything runs as `local`, so one set of registry rules serves both transports
     rather than one rule and an exception.
   - **Every credential variable is stripped from the processes this server creates** — by prefix,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,9 +46,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     serialised the server when the registry was global and one client could end another's targets;
     keeping it shared would have meant one client's four-minute pool walk making every other client
     wait for a boundary the registry now provides properly — namespaces that cannot be used
-    concurrently. Two connections presenting the *same* token still contend, which is one client
+    concurrently. A second *MCP session* for the same token still contends, which is one credential
     racing itself — and the `409` says so, where it used to report a second client that no longer
-    exists as a category.
+    exists as a category. Further requests carrying a session it already holds are not contention at
+    all and never were: they are served concurrently, which is what lets a `DELETE` arrive while a
+    tool call is still running.
   - Under **stdio** everything runs as `local`, so one set of registry rules serves both transports
     rather than one rule and an exception.
   - **Every credential variable is stripped from the processes this server creates** — by prefix,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     racing itself — and the `409` says so, where it used to report a second client that no longer
     exists as a category. Further requests carrying a session it already holds are not contention at
     all and never were: they are served concurrently, which is what lets a `DELETE` arrive while a
-    tool call is still running.
+    tool call is still running. The one `409` that means something else — a request arriving while
+    the sweeper releases this credential's expired sessions, when it holds nothing at all — now says
+    so, instead of describing a session the caller does not have.
   - Under **stdio** everything runs as `local`, so one set of registry rules serves both transports
     rather than one rule and an exception.
   - **Every credential variable is stripped from the processes this server creates** — by prefix,
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -377,9 +377,10 @@ so there is one set of rules and no transport exception. `docs/remote-listener.m
 half; what follows is what bites while editing.
 
 **Identity is ambient inside a call and by name outside one.** `crate::client::current()` reads a
-task-local that `listen::serve_one` sets around the whole MCP call, which is why no tool signature
-carries a caller. Anything running *outside* that scope — the listener's own diagnostics, a sweep, a
-shutdown — gets the default `local` instead of an error, so it must take the client as a parameter
+task-local that `listen::gate` sets — with `client::as_client`, around the `mcp.handle(req)` that
+is the whole MCP call — which is why no tool signature carries a caller. Anything running *outside*
+that scope — the listener's own diagnostics, a sweep, a shutdown — gets the default `local` instead
+of an error, so it must take the client as a parameter
 (`Sessions::live_count_for` against `Sessions::snapshot`). The bug this rule is written from was a
 log line reporting `local`'s session count to a named client on reconnect.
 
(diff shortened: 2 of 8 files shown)


glslang:
@coderabbitai review

glslang:
Fixed — and it is a fair catch on a PR whose entire point was that the words be right.

**Codex — describe MCP-session contention instead of TCP connections.** Correct. `admit` opens with
the arm that serves requests carrying the session this credential already holds, `in_flight` and
all, precisely so a `DELETE` can arrive on one connection while a tool call runs on another. So
"second connection" names something this server does not refuse, and an operator who reads it goes
looking for a network problem they do not have. I traded one wrong noun for another: the old text
said *client*, which stopped being true when tenancy became per client; mine said *connection*,
which was never true.

`Occupied` is a second **MCP session** for one credential — a fresh `initialize` while one is held
or being opened, or a request bearing an id that is not the one it holds. The body now reads
`this credential already holds a session here, or is opening one; one MCP session per credential at
a time`, and the same correction went through the five other places that repeated the phrasing: the
log line, the README, the smoke tier's test name and module header, and both prose descriptions in
`docs/smoke-test.md` and `docs/remote-listener.md`. The listener doc gained the sentence this
finding is made of — a `409` is never about connections, and here is what it *is* about.

The smoke tier's assertion moved with it: it matches on `one MCP session per credential` rather than
the old `one at a time`, so the needle is still the part of the sentence that carries the meaning.

ARM64 bench: clippy clean, 477 unit tests, dump-gated smoke tier 61 passed / 0 failed.

glslang:
@coderabbitai review

glslang:
Both fixed.

**Codex — point the identity guidance at the actual handler.** Right, and worse than a typo: a
section written to guide edits at the identity boundary is worth nothing if the call site it names
cannot be found. The handler is `listen::gate`, and the scope is `client::as_client` around the
`mcp.handle(req)` inside it. I have since checked every other symbol that section and the two
follow-ups cite against the source rather than trusting them — `client::strip_credentials`,
`Credentials::from_entries`, `kdconn::env_entries`, `Sessions::live_count_for`,
`Sessions::snapshot`, `Lease::admit`, `Lease::settle` — all present.

**Codex — describe MCP-session contention in the new follow-up.** Also right, and the sharper of the
two: item 28 exists to argue whether the gate should be *deleted*, so getting what it arbitrates
wrong is not a wording slip, it is arguing from a false premise. It now says `Occupied` is a second
MCP session for one credential, and says explicitly what it is not — concurrent requests on the held
session are what the gate lets through, `in_flight` and all.

The retirement argument is rewritten on that footing, and it is a better argument than the one it
replaces: not "clients earn `409`s for having connections" (they do not), but that `2026-07-28`
clients never become the holder at all, and that a client which lost its session id and
re-`initialize`s inside the grace is refused the adoption the identity now makes safe.

glslang:
@coderabbitai review

glslang:
Fixed, and it is the right catch on this branch specifically: making the message precise is what
exposed that one case does not fit it.

**Codex — preserve the teardown case in the conflict message.** Right. `admit` returns `Occupied`
while the sweeper is releasing this credential's expired sessions, and at that moment it holds
nothing and is opening nothing — so "already holds a session here, or is opening one" is not merely
vague there, it is the opposite of true. A client reading it goes looking for a session of its own
that does not exist.

It is now `Admission::Releasing` with its own body: *this credential's previous sessions are still
being released after its lease ran out; ask again in a moment.* A variant rather than a longer
sentence, because the two carry opposite advice — `Occupied` says waiting will not change anything,
and this one says it will, and a caller that cannot tell them apart cannot act on either.

`nothing_is_admitted_while_the_teardown_runs` covered this case and asserted `Occupied`; it now
asserts `Releasing`, which is the distinction it was always about — the server not being vacant yet
is not the same claim as this credential holding something.

ARM64 bench: clippy clean, 477 unit tests, dump-gated smoke tier 61 passed / 0 failed.

glslang:
@coderabbitai review

glslang:
Fixed, and following it up turned out to be worth more than the fix.

**Codex — account for stateless requests in the conflict message.** Right. A request with no session
id takes the opening path and holds the claim for its whole duration, so a second one from the same
credential lands on `Occupied` — and for a `2026-07-28` client that is *every* overlapping request,
because it never has an id to send. The message named only the second-session case, which for that
client describes something it cannot have. It now names both, with the remedy for each: send the
session id it holds, or ask again once the request in flight has finished.

**Then I measured it rather than reasoning about it**, and the measurement found something bigger,
filed as #168:

| request | result |
| --- | --- |
| `initialize` at `2026-07-28` | **200**, no `Mcp-Session-Id` minted — correct, SEP-2567 |
| `initialize` at `2025-06-18` | 200, mints one |
| `tools/list` at `2026-07-28`, no session id | **400** |
| same, with the discover-lifecycle `_meta` keys | **400** |

So the listener answers the newest handshake and then refuses the request after it. The server's own
log shows rmcp serving that handshake statelessly and correctly, which makes the next request the
interesting one. I have deliberately *not* chased it in this PR: the probe is hand-rolled and may be
tripping an earlier validator rather than the session rule, and settling it needs a client that
speaks the revision properly — which is what follow-up 29 asks for anyway. #168 has the evidence,
both candidate causes, and what is not yet known.

What did land here is the honest consequence: item 28 now says to settle #168 first (if the newest
revision is not served at all, what its clients do to the gate is the second question), and
`docs/remote-listener.md` tells an operator to negotiate a legacy revision until it is.

The overlapping-stateless-requests finding is recorded in item 28 as measured fact rather than
prediction — it is the strongest argument yet that the gate should not survive per-client ownership,
since `server_log` while a pool walk runs is exactly the overlap this repo documents as the way to
diagnose a wedged session.

ARM64 bench: clippy clean, 477 unit tests, dump-gated smoke tier 61 passed / 0 failed.

glslang:
@coderabbitai review

PR #180, opened 2026-08-21
Assert driver-frame attribution on an ARM64 stack (#154)

### LINKED ISSUE #154: Capture an ARM64 driver crash, so frame attribution is asserted on an ARM64 stack
#143 gave the debugger tier an ARM64 target and closed three of the four assertions that read one. The fourth — `a_driver_crash_names_the_driver_frame_that_analyze_cannot` — still runs against the x64 `MessageManager` dump on every architecture. That works (an engine with symbols reads it either way round, see #142), but it means **frame attribution to a third-party driver has never been exercised on an ARM64 stack**: the arithmetic that turns a captured frame into `module+RVA` off the load base is asserted only against x64 frames.

Closing that needs a crash, not a sample to find — and this bench is set up for one.

### What the bench already has (verified 2026-08-18)

On `Windows 11 Debuggee` (the ARM64 guest):

| | |
| --- | --- |
| driver | `C:\Users\goncalo\Desktop\HEVD\HEVD.sys` — PE machine `0xAA64`, a genuine **ARM64** build, 40,920 bytes, dated 2024-12-13, with `HEVD.pdb`, `HEVD.inf` and `hevd.cat` beside it |
| service | `HEVD`, `SYSTEM_START`, currently **STOPPED** with `WIN32_EXIT_CODE 31` (0x1F) — it is not loading |
| boot config | `bcdedit` shows `debug Yes` and **no `testsigning`**, which is the likely reason the driver will not load. First thing to fix |
| client | `C:\Users\goncalo\Desktop\hevd-exp\stack_buffer_overflow_token_stealing.exe` |
| dumps | already configured — `CrashDumpEnabled=3`, `AutoReboot=1`, minidumps to `C:\Windows\Minidump` |

### Why the sample we just checked in does not cover this

`docs/samples/121524-4703-01.dmp` is *already* an HEVD crash: its process is `stack_buffer_o…`, which is the client above, and its five siblings on that machine are the same run. But the bug check is `0xFC ATTEMPTED_EXECUTE_OF_NOEXECUTE_MEMORY` raised out of `nt!MiUserFault` — the kernel faulted **at the user-mode payload address**, after the stack was already redirected. So there is no `HEVD` frame anywhere on it, and its `faulting_frame` is a frame in no module at all: the payload page.

That is a genuinely useful fixture — it pins the in-no-module branch — and it is the wrong shape for attribution. The one this issue wants has HEVD itself on the stack.

### What the fixture has to show

Taken from what the x64 test asserts today:

- a `faulting_frame` that **exists**, below the top, under a stack of kernel internals a "blame frame 0" rule would name instead;
- that frame as `module+RVA` off the load base, with the RVA stable across load bases (the x64 one was identical in five dumps that loaded the driver at five different addresses);
- `!analyze` doing worse than the computed frame, which is why the frame is computed rather than taken from it;
- a process name, ideally over fifteen characters — the ARM64 sample already covers the *truncated* `_EPROCESS::ImageFileName` fallback, so this one could cover the audit-name path and the pair would cover both.

### Decisions before capturing

1. **Which HEVD path.** It has to bug-check *inside* the driver rather than after a redirect out of it — the token-stealing run above is precisely the case that leaves no driver frame. A corruption that faults on HEVD's own code path is what produces the stack this asserts on.
2. **PDB or no PDB.** `HEVD.pdb` ships beside the driver, so if it is on the debugger host's symbol path the frame resolves and `symbol` is *present* — the opposite of the x64 assertion, which pins `symbol` absent because `MessageManager` has none. Either keep the PDB off the path so the two tests make the same claim, or assert the symbolized form and let the pair cover both. Worth deciding before capturing, since it changes what the test says rather than how the dump is made.
3. **Size.** The ARM64 sample is 440 KB under the settings already on that machine; a second one of the same order is the budget.

### Capture notes

- The debuggee's kernel debugger link is **serial** to `WinArmSandboxDebugger` (`com:port=COM1,baud=115200`). With a debugger attached the bug check breaks in rather than writing the dump — continue (`g`) to let it write and reboot, or detach first with `kd -k com:port=COM1,baud=115200 -c ".time;qd"`.
- Parallels `Pause idle` must stay **off** on the debuggee, or a target sitting at a break is suspended by the hypervisor mid-run.
- Once captured, the test becomes an architecture-paired fixture like the other three rather than one dump named in the test body.

glslang:
Closes #154. `FOLLOWUPS.md` item 26.

## What was missing

The arithmetic that turns a captured frame into `module+RVA` off the load base had only ever run against **x64** frames. The ARM64 sample already checked in is a `0xFC` fault at a *user-mode* payload, so its faulting frame is in no module at all — the opposite shape.

## Getting a crash out of HEVD is the hard part

HEVD wraps its triggers in `__try/__except`, so every access violation it can raise comes back as a status:

| trigger | result |
| --- | --- |
| `NULL_POINTER_DEREFERENCE` | `STATUS_ACCESS_VIOLATION`, machine runs on |
| `BUFFER_OVERFLOW_NON_PAGED_POOL` | success; pool quietly corrupted |
| `ALLOCATE`/`FREE_UAF_OBJECT` twice | success twice; detected minutes later on a heap-maintenance worker thread as a `0x13A` with an **`nt`-only stack** — no driver frame anywhere |

What SEH cannot catch is a fail fast. `HEVD_IOCTL_BUFFER_OVERFLOW_STACK_GS` compiles its trigger with `/GS`, so overrunning the buffer corrupts the cookie and the driver's own `__report_gsfailure` runs `mov w0, #2; brk #0xf003`. The bug check is raised **inside the driver**, and the faulting frame is `HEVD+0x10dc` — that instruction's fixed offset in a fixed image.

`docs/samples/082126-7015-01.dmp`, 537 KB. The reproduction recipe is in `docs/smoke-test.md`; the IOCTL code was read out of the driver's own dispatch rather than taken from the widely-quoted HEVD list, which is wrong for this build (the code that list calls the null dereference is `FREE_UAF_OBJECT` here).

## Two decisions that went against what the issue proposed

**Not paired by architecture.** The issue proposed pairing, like the other three fixtures. The test is a table run on *every* host instead: an engine with symbols reads either dump either way round, so pairing would have meant an ARM64 runner stopped reading the x64 crash it reads today — trading coverage rather than adding it.

**The interesting disagreement is `!analyze`, not `symbol`.** The issue expected a choice about `HEVD.pdb` on the symbol path. What the capture showed is that `!analyze` *names* `HEVD`, where it calls the PDB-less `MessageManager` crash `Unknown_Module`. So the pair covers both: for one fixture the computed frame is the only thing naming the driver, for the other it is checked against an independent answer.

## Also: three bugs in `tools/ioctl_harness.ps1`

Found by using it. All Windows PowerShell 5.1-only, each fatal before an IOCTL was sent — em dashes in a BOM-less UTF-8 file, `0x80000000` as a negative `Int32`, and an empty `-InputHex` returning `$null`. A tool that advertises "no compiler needed" was unusable on a target whose only PowerShell is 5.1.

## Verification

On the ARM64 bench (engine bundle present, so nothing stands down):

- `cargo test` — 478 unit + 67 smoke, 0 failed
- `WINDBG_MCP_SMOKE_DUMP=1 cargo test --test mcp_smoke -- --nocapture` — **67 passed, 0 failed, no `SKIPPED` lines**
- `cargo clippy --all-targets` clean, `cargo fmt --all --check` clean


glslang:
**Depends on #179 for its ARM64 coverage.**

On this branch the ARM64 entry still stands every target read down — 7 `SKIPPED` lines, because `windows-11-arm`'s System32 has no `symsrv.dll` (that is exactly what #179 fixes). So on this PR alone, the new ARM64 fixture is asserted by the **x64** runner only.

That is worth stating plainly rather than leaving to be discovered, because the x64 entry reading an ARM64 dump was the one direction this repo had never verified — the bench measurement behind "either dump either way round" was an ARM64 host reading x64 dumps. It holds: `Smoke test (debugger tier)` here ran `a_driver_crash_names_the_driver_frame_an_all_kernel_walk_would_miss` with **no `SKIPPED` lines at all**, so `HEVD+0x10dc` was asserted on an x64 engine.

I have also run the two branches **merged** through CI (run 32456468158) rather than assuming they compose, since the combination is what lands on `main`.

PR #189, opened 2026-08-22
Change a service listener's clients without stopping it

glslang:
Closes `FOLLOWUPS.md` item 34.

## The problem

`--install-service` was the only writer of `%ProgramData%\windbg-mcp\token`, and the SCM refuses a second registration under the same name. So adding or rotating a client meant `--uninstall-service`, set every credential variable again, `--install-service`, `Start-Service` — which **drops every session the service holds**, a parked kernel attach included.

That was never a decision anyone made; it fell out of there being no writer but the installer. Two properties *were* chosen deliberately — "only the installer writes this file" and "never write through a file it did not create" — and neither of them requires the reinstall.

## What this adds

Three commands, from an elevated shell:

```pwsh
windbg-mcp.exe --add-listen-client    ci
windbg-mcp.exe --rotate-listen-client ci
windbg-mcp.exe --remove-listen-client ci
```

Each rewrites the credential file and then tells the running service to re-read it, **waiting for it to have done so** — so the set in force when the command returns is the set it wrote. Nothing is stopped and nothing is restarted.

**They generate the token and will not print it.** Standard output carries a fingerprint (`sha256:076C14953E1DE5EF`); the token goes to `<name>.token` beside the credential file. A token you typed has been through a shell history, and on a machine driven by an agent, through a transcript; one this server generated has been through neither. That is also what makes these narrow enough to allow-list in a permission rule, where "let this write `%ProgramData%`" is not.

**The operator does not name that path, and that is the fix rather than a limitation.** An earlier revision took `--token-out <path>`, which meant writing a live credential into a directory this program does not control the protection of: the ACL has to go on before the write (a DACL change does not revoke access through an already-open handle), doing that by path means create/close/ACL/reopen, and that hands anyone who can write the directory a window to substitute a file and keep a read handle to it. The state directory is already SYSTEM-and-Administrators-only with no traverse for anyone else, so there is no window to race and nothing to substitute.

## How the two hardened properties survive

- *Only the installer writes this file* becomes **only this program, running elevated, writes it** — the command is the same binary.
- *Never write through a file it did not create* survives as `service::write_credentials`: a fresh sibling created with `create_new` in the protected directory, ACL'd there, renamed over the old name. That also makes the replacement **atomic** for a service reading it concurrently. `finish_install` now shares that writer.

The ACL is untouched, and the environment is still read only by `--install-service`.

## The reload is the other half, not a footnote

Without it these commands would be an improvement in ergonomics and not in outcome, since a *restart* still costs the sessions. User-defined control code **128**, sent by the command and answered in `serve_as_service`'s handler, reaches `listen::reloaded` — which re-reads through the same `credentials()` that decides whether the listener may start at all (so a set that would not have started it cannot replace the one that did), swaps it into `client::Accepted`, and tears down a departed client.

The handler **blocks on the reload task's answer** (bounded at 10s, well inside the SCM's 30) and reports a failed re-read as a failed control code, so the command's printed claim is one it can back.

Three decisions worth reviewing on their merits:

- **A rotation keeps the name, and so keeps the sessions.** `Change` reports it as neither an arrival nor a departure — which is what stops the reload tearing down exactly the targets rotation exists to preserve.
- **A removal releases rather than refuses**, and closes the gate before it walks anything. A lease expiry needs no gate: it fires only after the client has been silent longer than any call can keep it quiet. A revocation has no quiet period — an opener that authenticated a moment earlier can be seconds from registering — so `Sessions::revoke` marks the owner under the registry lock and `admit` refuses it, the same shape `closing` gives shutdown. A session admitted behind a one-pass release would be a live kernel target owned by a client nothing can authenticate as.
- **Removing the last client is refused.** A listener with no credentials will not start, so that is a decision to stop serving rather than an incremental change — which is `--uninstall-service`.

A revocation whose reload could not be delivered **fails the command**: the credential you were taking out of service is still being accepted, which is not something to mention in passing. An `--add` in the same position is a warning. Two of these commands cannot run at once (`token.lock`, opened with no sharing): both would compute a whole file from their own snapshot and the later write would silently discard the earlier.

## Dependencies

CNG supplies both credential primitives — `BCryptGenRandom` for the token, `BCryptHash` against the SHA-256 pseudo-handle for the fingerprint — so this pulls in **no crypto crate**, just one `windows-sys` feature.

## Verification

`cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean, **496 unit + 70 smoke** tests pass on the ARM64 bench.

New coverage: token generation is held to the same presentability rule every configured credential is; a known-answer test for the digest; the fingerprint carries none of its token; `Accepted::replace` reports what moved and a rotation reports nothing; `Lease::revoked_for`/`forget` take a revoked client's sessions now and leave no entry a re-added name could inherit; a flag is never a value; the roster quotes no token; the reload code is one number both sides agree on; and a smoke test that a client command says what is missing **before** it opens the SCM or the file — which is what makes it safe to run on a host that has the service installed.

**Driven end to end against the installed service**, one process throughout, never restarted:

| step | result |
|---|---|
| `--add-listen-client bench` | log: `re-read the clients: added [bench], removed []`; the token answers `200` on an MCP `initialize` **the instant the command returns**; its file is `SYSTEM:Read \| Administrators:Read` in the protected directory |
| a second `--add-listen-client bench` | refused — the client exists |
| `--rotate-listen-client bench` | refused while the previous token file is still there ("would destroy the only copy"); succeeds once moved, old token `401`, new one `200` |
| `--remove-listen-client bench` | `401` immediately, and the log shows that client's MCP sessions being closed |
| `--remove-listen-client local` | refused — it was the only client left |
| a second command while the lock is held | refused, and writes nothing |

Across an earlier add/rotate/remove cycle the credential file came back **byte-for-byte** to its pre-test hash, with `local` untouched and the file returned from the JSON shape to the bare one.

## Docs

`docs/remote-listener.md`'s "Adding or rotating a client afterwards costs a reinstall" is replaced by the procedure that does not; the second foreground listener is re-framed as the development workflow it is rather than the fallback for this. The `windbg-debugging` skill's `setup.md` gains one clause and defers to that file. README and CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,9 +40,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   client is refused: a listener with no credentials will not start, and `--uninstall-service` is the
   command that means "stop serving".
 
-  A reload that cannot read the file **changes nothing** — the set is only ever replaced by one that
-  would have started this listener from cold, so a typo is a loud log line and a service still
-  serving, rather than every client locked out of a live kernel target.
+  The command **waits for the reload**, so the set in force when it returns is the set it wrote — the
+  control handler blocks on the reload task's answer and reports a failed re-read as a failed
+  control code. A reload that cannot read the file **changes nothing**: the set is only ever
+  replaced by one that would have started this listener from cold, so a typo is a loud log line and
+  a service still serving, rather than every client locked out of a live kernel target. That failure
+  is reported as an **error** for a `--remove` or `--rotate`, whose whole point is that a credential
+  stops being accepted, and as a warning for an `--add`.
+
+  Two of these commands cannot run at once (a `token.lock` in the state directory, opened with no
+  sharing): both would compute a whole file from their own snapshot and the later write would
+  silently discard the earlier.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   sharing): both would compute a whole file from their own snapshot and the later write would
   silently discard the earlier.
 
+  A removal also **deletes that client's token copy** if it is still sitting there: the credential
+  file is written by then, so the file authenticates nothing from the moment the command returns.
+
   A revocation **closes the gate before it walks anything** (`Sessions::revoke`). A lease expiry does
   not need to — it only fires after the client has been silent for longer than any call can keep it
   quiet, so nothing of that credential's can still be in flight — but a revocation has no quiet
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   out and the sweep that was to run on its next pass would not. Changing a token *without* losing
   the sessions is what `--rotate-listen-client` is for.
 
+  The registry gate a revocation closes is lifted **when the name is configured again**, not when
+  the teardown finishes. That release is one pass over the sessions that exist, so an opener which
+  authenticated before the revocation and has not registered yet — an `attach_kernel` is a worker
+  spawn away — is invisible to it, and lifting on "nothing left to release" let that opener register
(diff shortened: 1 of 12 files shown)


glslang:
@coderabbitai review

PR #196, opened 2026-08-23
Serve a surface per client, not per server

glslang:
Closes `FOLLOWUPS.md` item 36.

`--tools` ([#195](https://github.com/glslang/windbg-mcp/pull/195)) narrowed what a run advertises, and it narrowed it for *everybody* on that run. That is right for stdio, which has one client by construction. It is the wrong shape for what the listener exists for: a local model that can hold twenty tools and a hosted client that can hold fifty-one, pointed at the same box and the same debug sessions, told apart by their bearer tokens already.

So a client may now be configured with a spec of its own — `WINDBG_MCP_TOOLS_<NAME>` beside its token, or a `tools` field in a credential-file entry that is then an object rather than a string:

```pwsh
setx WINDBG_MCP_LISTEN_TOKEN_BENCH "<a long random string>"
setx WINDBG_MCP_TOOLS_BENCH        "session,inspect,crash"
```

```json
{
  "local": "<a long random string>",
  "bench": { "token": "<another>", "tools": "session,inspect,crash" }
}
```

Two credentials on one port get two `tools/list` answers.

## The three decisions

**The run's `--tools` is the default, not a ceiling.** A client's own spec replaces it, wider or narrower, rather than intersecting with it: an intersection can produce a surface neither the operator nor the client ever named, while "which of these two is in force" has an answer either way. `session` is added to a client's spec exactly as it is to a run's.

**A spec is parsed where a credential is validated.** `Credentials::build` runs it through the flag's own parser, so a surface this server could not serve is refused at startup rather than discovered by a caller whose tool list came back empty. `Toolset::parse_from` takes the source's rendered name — the vocabulary is written down in three places now, and a refusal that always said `--tools` would send an operator to a command line they never typed. A spec naming a client nothing configures a token for is refused too, on the precedent of the two collisions that configuration already refuses: it is a setting that could never take effect, and the way to write one is the typo that makes `WINDBG_MCP_TOOLS_BENCH` and `WINDBG_MCP_LISTEN_TOKEN_BENCH` disagree.

**Nothing sends `notifications/tools/list_changed`.** This server keeps no peer handle to notify an MCP session through, and `2026-07-28` has no session to notify at all, so it would be a guarantee on one revision and silence on the other. A surface is instead fixed where the caller is identified — `initialize`, or every request on the stateless revision — so a change reaches a client when it next connects, which is one sentence for both revisions. `--set-listen-client-tools` says so where an operator reads it.

## Two things this had to correct rather than add

- The refusal for a tool off the surface told **every** caller to widen `--tools`, which is the wrong command for a client with an entry of its own. It now names whichever configuration chose the surface (`Toolset::refusal`, `Chosen`).
- `--add-listen-client` printed *"it gets the whole tool surface, as every client here does"*, which this makes false. It says what the client is actually served.

## A fourth client command

`--set-listen-client-tools <name> --tools <spec>` changes a service-hosted client's surface without a reinstall or a restart, the way item 34's three commands change a token; with no `--tools` at all the client goes back to the service's own surface. It mints no credential and revokes none. A `--tools` beside `--rotate-` or `--remove-listen-client` is **refused** rather than ignored, because it reads exactly like a command that had narrowed that client.

## Coverage

The surface is resolved on the line that captures the caller's identity — which was wrong for two months and right in every unit test (item 29). So the assertion that matters is two tokens on one listener seeing two different `tools/list` answers, on the session-bearing route *and* the stateless one: `mcp_smoke::two_clients_on_one_listener_are_served_two_surfaces`. Protocol tier, because a tool list and a surface refusal need no debug target. Beside it: the env/file configuration paths, the orphan and collision refusals, the file's new object shape and its round trip through the installer's writer, and both halves of the refusal message.

Verified on the ARM64 bench (tree hash-matched to the branch first): `cargo clippy --all-targets -- -D warnings` clean, **526** unit tests, **75** smoke tests, and the debugger tier with `WINDBG_MCP_SMOKE_DUMP=1` (69.5 s, so it ran). `cargo fmt --all --check` clean.

Docs: README, `docs/remote-listener.md` (new *A tool surface per client* section, the file's object shape, the fourth command), `docs/local-model.md`, `docs/token-budget.md`, `CLAUDE.md`, `CHANGELOG.md`, `FOLLOWUPS.md`, and the setup skill. One stale claim corrected along the way: `local-model.md` still said a service-hosted listener needs a reinstall for a second client, which item 34 fixed.

## Summary by CodeRabbit

* **New Features**
  * Configure a distinct tool surface for each listener client using environment variables or credential files.
  * Add or update client tool settings without restarting the service; changes apply on the next connection.
  * Manage client tools through service commands, including assigning, changing, and clearing tool sets.
  * Display each client’s effective tool surface and provide clearer guidance when a requested tool is unavailable.

* **Documentation**
  * Updated listener, budgeting, setup, and local-model guidance to explain per-client tool configuration and precedence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,9 +34,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and revokes none. A `--tools` beside `--rotate-` or `--remove-listen-client` is refused rather
   than ignored, because it reads exactly like a command that had narrowed that client.
 
-  **A surface change reaches a client on its next connection**, and nothing announces one. The
-  surface is fixed where the caller is identified — `initialize` for a client holding an MCP
-  session, every request for one on `2026-07-28` — and no `notifications/tools/list_changed` is
+  **A surface change reaches a client the next time it is identified**, and nothing announces one.
+  The surface is fixed at that moment — `initialize` for a client holding an MCP session, every
+  request for one on `2026-07-28`, which therefore picks the change up with nothing done to it —
+  and no `notifications/tools/list_changed` is
   sent: this server keeps no handle to notify a session through, and the sessionless revision has
   no session to notify, so it would be a guarantee on one revision and silence on the other. The
   command that changes a surface says so.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,42 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **A tool surface per client, not just per server** (`FOLLOWUPS.md` item 36, which closes it). A
+  `--listen` server names its clients already, and they do not have one budget between them: the
+  arrangement this listener exists for is a local model that can hold twenty tools beside a hosted
+  client that can hold fifty-one, on the same box and against the same debug sessions. So a client
+  may be configured with a `--tools` spec of its own — `WINDBG_MCP_TOOLS_<NAME>` beside its token,
+  or a `tools` field in the credential file — and two credentials on one port get two `tools/list`
+  answers.
+
+  **The run's `--tools` is the default rather than a ceiling.** A client with no spec of its own is
+  served whatever the run serves; a client with one is served that instead, wider or narrower,
+  because an intersection would produce a surface neither the operator nor the client ever named.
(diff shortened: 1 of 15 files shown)


glslang:
@coderabbitai review

PR #205, opened 2026-08-23
Run the surface, the window and the model as a grid

glslang:
Follow-up to [#196](https://github.com/glslang/windbg-mcp/pull/196), which made the tool surface a property of the *client*. That is the knob this uses: three surfaces are three credentials on one listener, so a cell changes the bearer token and nothing else — the feature under measurement is the shipped one rather than a test double.

`docs/local-model.md` had two runs of one model, on one surface, at one window. Both say a 51-tool surface is drivable; neither can say which knob carried the result. This runs the grid: **3 local models × 3 tool surfaces × 3 context windows, plus Claude Code as the control — 33 cells, 150 task runs, 80 minutes.**

## What it found

**The window was not the binding constraint.** At a *served* 8,192-token window a 17,300-token surface answered all six tasks correctly, including three multi-turn tasks and two carrying ~10,000 characters of tool output. The arithmetic in `local-model.md` predicted this could not work; it works. That page now says so and points at the eval.

**The surface axis costs fewer answers than tools.** 51 → 11 removes 40 tools and 2 of 6 answers, because `open_dump`'s summary carries the bug check *and* the module count, and `crash_triage`'s frame 0 is the `pc` that `registers` reports. What narrowing does produce is calls to tools that are not there — from **every** model including both control rows, and **never** from a 51-tool cell. The refusal that names the client's own surface is load-bearing.

**Failure modes separate the models more than scores do.** Same surface, same two tools, same `crash_triage` output: qwen derives the `pc` from frame 0, gemma asks four times for `debug_batch` (not served) and returns an empty answer, nemotron reports bug check parameter 1 as the `pc` with no hedging. Opus makes nemotron's mistake here — which is what a control is for.

Scores at 262k, correct out of answerable: Opus 15/15, Sonnet 15/15, qwen3.8:27b **15/15**, gemma4:31b 14/15, nemotron-3.5:30b 10/15.

## What is new

| | |
| --- | --- |
| `tools/local_model_eval.py` | the matrix runner and the grader (`--grade`, `--matrix`) |
| `tools/claude_code_drive.py` | the control row: `claude -p` through the same listener, same allow-list |
| `tools/eval_tasks.json` | six tasks and the answer key, read off the checked-in dumps with this server's own tools |
| `tools/bench_listener.ps1` | one foreground listener, three clients, three surfaces, tokens on stdin |
| `docs/local-model-eval.md` | the method and the results |

`tools/local_model_drive.py` grows only what a grid needs: `NUM_CTX`, a JSON record per task, and a three-way verdict per call.

## Two properties that are easy to lose

**Record what the runtime *served*, not what you asked for.** `num_ctx` does not shrink an instance ollama already holds — five cells asking for 8,192 were served 32,768 and looked perfectly healthy, which is exactly the result the context axis exists to find and would have been fiction. `/api/ps` was the only disagreement. The runner now evicts between windows, every record carries `served_context`, and the grader marks a cell where the two differ. Those five were dropped and re-run.

**The Claude row needs four fences or it measures something else**: `--strict-mcp-config` (or it falls back to the editor's registered server, a different credential with the whole surface), `--disallowedTools` for the built-ins (or it answers from this repository), `ENABLE_TOOL_SEARCH=false` (or its MCP schemas are deferred and the surface costs it nothing while every other row pays), and a neutral working directory. Its prompt-token column is not comparable with the ollama rows and the document says so rather than quoting it.

## Corrections, recorded rather than smoothed over

Three grader bugs, each of which had already produced a wrong verdict: `0x22` matching inside `0x22200B`; `0x0802` failing to match `0x802`; `0xfffff801_3c65bca8` failing to match the same address without its separator. **Two of the three were caught by reading the control's answers.** One answer-key prediction was wrong — `arm64_pc` is answerable on every surface, verified against the dump before the key changed. Grading runs over the whole log at the end, so every cell is scored by the same final key.

## Verification

`markdownlint-cli2@0.23.2` clean over CI's globs. No Rust changed, so no `cargo` gates move. The grid itself is the verification of the harness: 33 cells, resumed once mid-run after the eviction fix, with 0 mis-served cells in the final log.

`FOLLOWUPS.md` item 39 has what this does not cover: every conversation here is short, so the case where the transcript rather than the surface fills the window is untested, and grading an investigation needs a different key.

## Summary by CodeRabbit

- **New Features**
  - Added a comprehensive local-model evaluation workflow covering models, tool surfaces, context windows, answer grading and benchmark reporting.
  - Added automated evaluation support for Claude Code and local model runs, including task isolation, resumable results and performance metrics.
  - Added benchmark listener support for configuring and launching different tool-surface profiles.

- **Documentation**
  - Added evaluation findings, limitations, methodology and usage guidance.
  - Updated project references with links to evaluation, disassembler coordinate and smoke-test documentation.
  - Recorded follow-up opportunities for multi-step investigations and mutating-tool evaluation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ package-lock.json
 
 # Local Claude Code config
 .claude/
+
+# Output of the local-model eval (`tools/local_model_eval.py`), which the checked-in plan
+# writes here: run logs and one JSON record per task. Measurements, not sources.
+eval-out/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   listener that serves all three surfaces — which is 0.11.0's per-client `--tools` doing the
   narrowing, rather than a test double.
 
+  It also found a defect in this server, filed as `FOLLOWUPS.md` item 40: **`--tools` narrows
+  `tools/list` and not the `instructions` string**, which is one compile-time constant naming
+  twenty-one tools and is sent to every client whatever its surface. A `crash`-surface client is
+  served 11 tools and told about 21, of which 17 it cannot call - which is where every "invented"
+  tool call in the grid came from, and 59% of a 497-token string that client pays for in full.
+
   Three findings worth the run. **The window was not the binding constraint**: at a *served*
   8,192-token window a 17,300-token surface answered all six tasks correctly, multi-turn tasks with
   10,000-character results included, so the "will it fit" arithmetic in `docs/local-model.md`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **The local-model eval: a grid where there were two sightings**
+  ([`docs/local-model-eval.md`](./docs/local-model-eval.md), `FOLLOWUPS.md` item 39). Three ~30B
+  local models against three tool surfaces and three context windows, with Claude Code as the
+  control, scored against an answer key read off the checked-in sample dumps with this server's own
+  tools before any model saw them. `tools/local_model_eval.py` runs the grid and grades it,
+  `tools/claude_code_drive.py` is the control row, `tools/bench_listener.ps1` stands up the one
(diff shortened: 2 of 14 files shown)


glslang:
@coderabbitai review

PR #206, opened 2026-08-23
Tell a client about the tools it is served, and no others

glslang:
Stacked on [#205](https://github.com/glslang/windbg-mcp/pull/205), whose eval found this and whose `FOLLOWUPS.md` item 40 describes it. **Base is `local-model-eval`, so merge that first.**

`--tools` narrows the **router** per client: `tools/list` answers with that client's set, and a call for anything else is refused by name. The `instructions` string sent at `initialize` never moved with it — one literal on `#[rmcp::tool_handler]`, naming twenty-one tools, sent to everybody:

```
min   served 11 tools | instructions 1990 chars | names 21 tools, 17 of them not served
full  served 51 tools | instructions 1990 chars | names 21 tools,  0 not served
```

## What it cost, measured rather than assumed

**Wasted turns.** Every off-surface call in #205's grid is one of the seventeen names this string advertises — both Claude control rows asked for `modules` and `execute`, and gemma spent an entire task's turn budget re-asking for `debug_batch`. The eval recorded that as models *inventing* tool names; they were reading this server's own advertising, and its metric is renamed `unserved` in that PR because of this one.

**Context, on the surface least able to pay it.** 1,990 characters, ~497 tokens, identical for every client, of which **59% was sentences naming only tools a `crash` client cannot call** — about 12% of that client's whole prompt. `--tools crash` dropped 54,000 bytes of schemas and kept every word of the prose selling what it dropped.

## The shape

A base plus a fragment per **group**, assembled for the client's own `Toolset` — not a filter over the old text, because that text names several tools per sentence and cutting by name leaves mangled English in the first thing a model reads. `Toolset::serves_group` is the question each fragment asks, and it is *any* rather than *all*: a spec may name one tool out of a family (`--tools registers`), and that client still needs the sentence saying what the family is for.

`get_info` is now written out, which the macro yields to exactly as it does for `call_tool` (`// Auto-generate get_info() if not already provided`). **`name` had to move with it** — left on the attribute it is only read while generating the `get_info` that no longer exists, and the server would introduce itself as `rmcp` at the SDK's version, which a protocol-tier assertion catches.

| Surface | Instructions |
| --- | --- |
| all 51 tools | 1,983 (the constant it replaces was 1,990) |
| `session,inspect,crash` | 1,220 |
| `crash` | **927** |

The allocator family stays absent as it was before: this is about not advertising what a client cannot call, not about advertising more.

## Coverage

The per-client property is asserted the only way it can be — **two credentials on one listener**, extending `two_clients_on_one_listener_are_served_two_surfaces`: `bench`, served `crash`, must not read the words `modules`, `execute` or `debug_batch`; `local`, served `session,inspect`, must not read `crash_triage`, `debug_batch`, `decode_ioctl`, `ttd_calls` or `run_to_address`; both must still carry the base. With one credential, "this client's instructions" and "the run's" are the same string — the same reason that test exists at all.

The budget golden moves by exactly seven characters (`instructions 1990 -> 1983`) and nothing else: no tool's schema, description or annotations changed, which is what reading that diff is for.

## Verified on the ARM64 bench

`cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean, **530** unit tests, **76** smoke tests, and the debugger tier with `WINDBG_MCP_SMOKE_DUMP=1` at **63.7s**, so it ran rather than skipped.

## Summary by CodeRabbit

- **Bug Fixes**
  - Client instructions now reflect only the tools available on that client’s configured tool surface.
  - Removed references to unavailable tools, reducing confusion and prompt size for narrowed tool sets.
  - Server information now consistently includes the correct identity and tailored guidance.

- **Documentation**
  - Updated changelog and guidance on local model evaluation and token budgets.
  - Added measured instruction sizes for full and reduced tool surfaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A client is told about the tools it is served, and no others** (`FOLLOWUPS.md` item 40, which
+  closes it). `--tools` narrowed the router per client — `tools/list` answers with that client's
+  set, and anything else is refused by name — while the `instructions` string sent at `initialize`
+  stayed one literal naming twenty-one tools, sent to everybody. A `--tools crash` client could
+  call eleven tools and was told about twenty-one, so it would ask for `modules`, `execute` or
+  `debug_batch` and be refused; driving this server with local models measured every one of those
+  wasted calls (`docs/local-model-eval.md`) and read them as models inventing tool names. They were
+  reading this server. The string is now a base plus a fragment per group, assembled for the
+  client's own surface: 1,983 characters for the whole surface against the constant's 1,990, 1,220
+  for `session,inspect,crash`, and **927 for `crash`** — which also removes ~12% of that client's
+  prompt, on the surface that exists to be small.
+
 - **`--list-listen-clients` sorts both halves of what it prints.** A credential file's entries
   already came back sorted; the environment's arrived in the order the variables were scanned,
   which on Windows is by *variable* name — so `WINDBG_MCP_LISTEN_TOKEN` came before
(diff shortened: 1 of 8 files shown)


glslang:
@coderabbitai review

PR #239, opened 2026-08-26
Open a 32-bit .NET dump with an engine that can load its SOS

### LINKED ISSUE #234: Support loading SOS for 32-bit .NET Framework dumps (x86 engine worker or guidance)
## Problem

When a 32-bit .NET Framework process dump is opened in windbg-mcp, SOS cannot be loaded. The engine worker is x64, and the two available SOS DLLs both fail:

- **64-bit SOS** (`Framework64\v4.0.30319\sos.dll`) loads into the x64 worker process but refuses to analyze the x86 target: `SOS does not support the current target architecture (14c)`
- **32-bit SOS** (`Framework\v4.0.30319\sos.dll`) can't load at all: `Win32 error 0n193 "%1 is not a valid Win32 application"`

`.effmach x86` is already set (the engine recognizes the dump as x86), but that only affects how the engine interprets the target, not which DLLs the host process can load. `!wow64exts.sw` doesn't apply either, because the dump is a pure 32-bit capture (no WoW64 layer present).

The result: `!threads`, `!clrstack`, `!gcroot`, `!dumpheap` and the rest of the SOS surface are unreachable. Native analysis (stacks, modules, memory reads) works fine.

## Context

These dumps come from 32-bit Windows services running under WoW64, captured with the 32-bit procdump (`procdump.exe -ma <process>`). They're common in environments running legacy .NET Framework 4.x services that haven't been migrated to 64-bit or .NET 8+.

## Workarounds I found

**For future dumps:** capture with the 64-bit procdump (`procdump64.exe -ma <process>`). That produces a dump containing the full WoW64 layer, which the x64 engine + x64 SOS + `!wow64exts.sw` can handle.

**For existing 32-bit dumps:** copy the x86 `cdb.exe` and its DLLs out of the WinDbg Store package and run it separately:

```powershell
$pkg = Get-AppxPackage Microsoft.WinDbg
Copy-Item (Join-Path $pkg.InstallLocation "x86") "$env:TEMP\cdb32" -Recurse -Force
& "$env:TEMP\cdb32\cdb.exe" -z "dump.dmp" -c ".loadby sos clr; !threads; ~*e !clrstack; q"
```

This works (the 32-bit host loads the 32-bit SOS without issue), but it's outside windbg-mcp's session, so the AI agent can't drive it.

## Possible approaches

A few options, in decreasing order of ambition:

1. **Spawn an x86 engine worker** when the dump is detected as x86 + managed. The x86 `cdb.exe`/`dbgeng.dll` ships alongside the x64 one in the WinDbg package. The worker could be the x86 binary for that session, transparent to the caller.

2. **Shell out to x86 cdb for SOS commands** when the architecture mismatch is detected. Run the SOS command in a child x86 cdb process and return its output through the existing tool surface.

3. **Document the limitation and the workaround** in the crash-dump skill or README, so an agent knows to tell the user about `procdump64.exe` for future captures and falls back to the x86 cdb workaround for existing ones.

## Environment

- windbg-mcp 0.12.0 (scoop, dungeon bucket)
- WinDbg Preview (Microsoft.WinDbg) from Microsoft Store
- Windows 11 x64
- Target: 32-bit .NET Framework 4.8 service dump, captured with 32-bit procdump

glslang:
Closes #234.

A 32-bit .NET Framework dump can now be opened and its SOS extension used — `!sos.threads`, `!clrstack`, `!dumpheap` and the rest — by putting the engine in a process of the target's architecture and driving it over DbgEng's remote transport.

## Why the engine has to move

An extension DLL is loaded into the debugger's **own** process, so its architecture is the host's, not the target's. Measured on Windows 11 x64 against a 32-bit managed target:

- 32-bit `sos.dll` into an x64 host → `Win32 error 0n193`
- 64-bit `sos.dll` loads, then → `Failed to load data access DLL, 0x80004005`

The second is what closes the door: `mscordacwks` is paired to the **target's** architecture as well as the host's, so there is no SOS + DAC combination an x64 host can use against a 32-bit CLR. That rules out every in-process fix, including ClrMD/`dotnet-dump`, and `dbgsrv.exe` (a process server proxies target access but leaves the engine local, so SOS stays x64). Moving the engine is also what WinDbg itself does — its package ships `EngHost.exe` under both `amd64\` and `x86\`.

## How it works

`src/dump.rs` reads the dump's architecture from its own header, `src/enginehost.rs` starts an `x86\cdb.exe` as a debugging server, and dbgscope's new `DebugEngine::connect` (glslang/dbgscope#119) drives it.

**The decision has to precede the engine.** `GetEffectiveProcessorType` would answer authoritatively, but only once a session exists in a process whose architecture is by then fixed — the thing being chosen. It also cannot be revisited later: `Ready` means the engine exists, and `INTERRUPT` is taken from it once into a `OnceLock`, so an engine swapped mid-session leaves `interrupt` pointing at a dead one. Hence a header read, a few hundred bytes off a file that is routinely hundreds of megabytes.

**The worker decides; the supervisor only forwards the path.** The supervisor forwards it because it holds the opener op before it spawns anything. The worker decides because keeping the architecture read, the host search, the fallback and the explanation in one process is what stops a caller being told half a story from each.

**A failure to place the engine elsewhere is not a failure to open.** An x86 dump opens today and its native analysis works; that keeps working where there is no x86 `cdb.exe`, or where starting one fails. Those paths fall back to this process and carry a `limitation` saying SOS is unreachable and what to copy — on **both** halves of the result, since a structured-aware client forwards `structuredContent` and drops the text.

## Cost

`limitation` adds 40 bytes of `outputSchema` per opener and **nothing the model reads** — the golden's `modelVisible` column does not move. No new tool, so no `toolset.rs` group and no surface change.

## Known gap

`modules` rows carry no `pdb` on a remote session: `IDebugAdvanced2::GetSymbolInformation` does not cross the transport. `E_INVALIDARG` invites the reading "the caller passed a bad `SizeOfStruct`", so:

| Transport | Engine arch | `module_pdb` |
| --- | --- | --- |
| in-process | x64 | **pass** |
| remote npipe | x86 | fail `E_INVALIDARG` |
| remote npipe | x64 | fail `E_INVALIDARG` |

Rows 1 and 3 differ only in the transport, same dump, same module, same base, same `Pdb` symbol state. `with_pdb_identity` already drops the field rather than failing the listing. `FOLLOWUPS.md` item 49 records the fix worth making — parse GUID/age from the image debug directory, which closes it for every transport.

## Testing

- 575 unit tests, 84 smoke tests, both tiers enabled.
- New **32-bit managed dump** tier (`WINDBG_MCP_X86_DUMP=<path>`), asserted through the tool surface against a real 32-bit `procdump -ma` capture.
- The tier asserts `!sos.threads` **module-qualified** — not incidental. `open_dump` loads `ext.dll`, whose own `!threads` answers on any engine; the first version of this test passed a perfectly good native thread table off as proof.
- The fixture is gated on a path rather than a flag because a full-memory capture is several times this repository. Both bitnesses of procdump were measured: the 32-bit capture reads as x86 and the 64-bit capture of the *same process* reads as x64, which is the discrimination the routing turns on.

## Also in here

`FOLLOWUPS.md` item **50** is unrelated to #234 and was found while testing it: Defender quarantined this project's own binary as `Trojan:Win32/Bearfoos.B!ml`. The released binary has the same profile — `FileVersion`, `CompanyName` and `ProductName` are all empty, and the Sigstore attestation is a supply-chain claim nothing on the machine reads rather than Authenticode. `microsoft/apm#487` is the same detection on Microsoft's own shipped binary.

`CLAUDE.md` also gains a correction it earned: the engine bundle belongs in `target\debug\deps` as well, which is where libtest's binaries run from. Reading it as `target\debug` alone cost a test that failed with an error about something else entirely.

## Summary by CodeRabbit

* **New Features**
  * Added support for analysing 32-bit .NET dumps through a compatible x86 debugging engine.
  * Sessions now report clear limitations when required x86 support is unavailable.
  * Added architecture-aware dump detection and optional 32-bit smoke testing.

* **Documentation**
  * Expanded setup, structured-results, and troubleshooting guidance for 32-bit dump analysis.

* **Tests**
  * Added coverage for dump architecture detection, engine hosting, and 32-bit managed dump workflows.


glslang:
@coderabbitai review

Issue #226, opened 2026-08-25
execute silently swallows execution-control commands (g/p/t) — a plain Execute sets run state without pumping, and every later control call fails with 0x80040205


daiaji:
## Summary

The `execute` tool routes through `EngineOp::BoundedCommand` → `raw_command` → a plain `IDebugControl::Execute`. This repo's own docs say what that means (docs/tool-surface.md, "Control flow and the TTD wrappers"):

> a plain `Execute` only sets the run state — it doesn't move the target

Yet `execute` performs **no content check whatsoever**: it happily accepts `g`, `p`, `t`, compound commands like `bp X; g`, and breakpoints with embedded command strings. When any of these is issued through `execute` against a live target:

1. The call returns the echoed command text as if it had run — no error, no output (a silent swallow).
2. The target does not move.
3. **Every subsequent execution-control call on that session fails with `0x80040205`** ("An unexpected exception was raised") for the rest of the session's life. Read-only commands (`bl`, `.bpcmds`, `.lastevent`, `threads`, disassembly) keep working, which makes the session look half-alive.

There is no recovery short of `end_session`.

## Reproduction (controlled, n=3 isolated)

Live user-mode target, launch → initial int3:

1. `set_breakpoint` (typed tool) at `module+RVA` ✓
2. `go` (typed tool) → hits the breakpoint normally ✓
3. `registers` ✓
4. `execute("g")` ← the single variable introduced — returns only the echoed command text
5. `go` again → `Debug command failed: An unexpected exception was raised (0x80040205)`; every later control request fails the same way

Variants confirmed in the same campaign:

- `execute("bp X; g")` — the `bp` half takes effect (visible in `bl` afterwards), the `g` half never runs; the session wedges as above.
- A breakpoint armed via `execute` with an embedded command string (`bp <addr> ".printf \"…\\n\"; gc"`) — `.bpcmds` reads the string back character-perfect (transport is not corrupting anything), but the first `go` after setting it dies with `0x8000FFFF`, and from there the session is wedged as above.

A no-execute control group (launch → `set_breakpoint` → `go` ×3 → `registers` → `bc` → close) completed cleanly across every run. So neither breakpoint count nor repeated `go` calls trigger any of this — the poison is execution-control text routed through `execute`.

One diagnostic trap worth flagging: a crash-dump session refuses `set_breakpoint` with the *same* `0x80040205` (dumps have no live thread to arm int3 on), so the code alone does not tell these two situations apart.

[...]

PR #243, opened 2026-08-26
A target that ends is an ending, not a catastrophe (#242)

### LINKED ISSUE #242: A target exiting during the resume pump leaves the session half-dead and can take the worker down
## Summary

On launch/attach_process sessions, when the target process exits while the engine is running or being pumped to a stop after a resume, the session enters a broken state:

- the command whose release raced the exit returns only a silent echo; the pump's captured output (breakpoint banner, embedded script output) is discarded,
- thread-dependent commands (`k`, `registers`) then fail with a bare `E_UNEXPECTED` (0x80040205) while `.echo` and `.lastevent` still answer, which is indistinguishable from a wedged session,
- execution control issued in that state is the unguarded half of the hazard `execute_and_wait` refuses up front (item 2 under "The part only you can land"); on this machine the worker died in that race once.

Expected: the resume reports "the target exited" and later commands answer a clean no-debuggee-class error. Actual: silent echo, then the chain above.

Related: [#226](https://github.com/glslang/windbg-mcp/issues/226) resolved the unpumped-resume case; this is the exit-during-pump corner it left open, and the unguarded raw-command AV path.

## Root cause

The exit races the pump. The engine is left with no current process but can still report a stale live status (observed `0x7` `DEBUG_STATUS_GO_NOT_HANDLED`; alternatively `DEBUG_STATUS_NO_DEBUGGEE`), and `GetLastEventInformation` answers "no event". `settle` only asks whether the engine is running, so a pump that ended with the target's exit is classified as "fine" (`is_running()` -> false) and its captured output is dropped.

## Repro

Minimal, no breakpoint and no symbols needed:

```text
launch <any short-lived process, e.g. cmd.exe /c ping -n 2 127.0.0.1>   # stops at the initial breakpoint
execute 'g'                                                             # the process exits while the engine pumps
execute 'k 3'                                                           # raw 0x80040205
```

The same failure follows any resume that races the exit:

1. `execute 'bp <hot> "du @rdx L40; .echo ====; g"; g'` — the breakpoint hits (verified: `Breakpoint 0 hit`), the embedded `g` auto-continues, the process exits mid-pump;
2. `execute 'g; g; g; g; g'` after a first release — the later resumes hit an engine that is already running or has already lost the target, and fail with 0x8000FFFF;
3. `execute 'bp <never-hit>; g'` — nothing stops the engine, the process runs out and exits.

Timing: the fault requires the target to exit during the pump, so a short-lived debuggee triggers it and a long-running one never does. The worker-death variant is a race and intermittent; run the sequence 2-3 times before treating a clean run as fixed.

## The part only you can land (dbgscope)

The fixes below live in dbgscope, whose pin only you can move. Everything else is already handled on our side (see below), so once these are in, this issue is done.

1. `DebugEngine::settle`: when the pump ends because the target exited, return a terminal `CommandRun` that carries the captured output. Currently the wait ends with `S_FALSE`/an error and `settle` returns `Err` with the buffer discarded, so an embedded `bp X "du @rdx L40; g"` script's output is lost precisely when the process exits.
2. `execute_command_bounded` (raw `Execute`): apply the same `DEBUG_STATUS_NO_DEBUGGEE` refusal that `execute_and_wait` has, so execution-control text never reaches an engine with no debuggee. dbgscope documents that this case can push DbgEng into an access violation that tears down the whole process; the guard belongs in the primitive, not in each caller.
3. Optional: `settle` could state whether it pumped to a stop or to an exit, and engine script-error narration could be structured (e.g. "script segment N failed") so a raw HRESULT inside command output is not mistaken for a server failure.

## Not part of this ask

- The `Debug command failed: ... (0x8000FFFF)` text that a second resume in one command string prints is DbgEng correctly refusing an illegal operation; it should stay.
- Text-level pre-scanning of command strings to refuse multi-resume scripts is out of scope: the codebase's design rejects command-name lists as non-exhaustive, since aliases, `.if` branches and the data model all reach execution.

## Already handled on our side

A worker-side mitigation is in place in our local build: once the engine reports no debuggee on a launch/attach session, the session is marked terminated, the observing command carries an explanatory note, later commands are refused with the reason, and `EndSession` stays callable. It closes the half-dead chain and the AV window for commands issued after the exit, but it does not recover the pump's captured output and does not guard the in-script second-resume case — those are items 1 and 2 above. We are happy to share the patch if you want it.

## Acceptance

- `settle` returns a terminal `CommandRun` with the captured output when the target exits during the pump; after the minimal repro, `k 3` answers a clean no-debuggee-class error instead of the `E_UNEXPECTED` chain; no worker death across repeated runs of the race.

Environment: Windows 10 x64 (10.0.19045), crate 0.12.1 (2fc8823), dbgscope 19a87d8.

glslang:
Closes #242, and with it [`FOLLOWUPS.md` item 48](https://github.com/glslang/windbg-mcp/blob/main/FOLLOWUPS.md), which had held the same question open since #226: *is an exited target an error at all?* It is not.

> [!IMPORTANT]
> Depends on **glslang/dbgscope#120**. The `rev` here points at that branch's head; it needs repointing to the merge commit before this lands.

## What was wrong

Everything below was measured on dbgeng 10.0.26100.1 (ARM64) before anything was written. Two of the issue's premises did not survive that, and both changed the fix.

**A target running to completion was a catastrophe.** `WaitForEvent` answers `E_UNEXPECTED` once the target is gone, and that was propagated verbatim — `Debug command failed: Catastrophic failure (0x8000FFFF)` for a program exiting normally — with the buffer the run had captured discarded alongside it. That buffer is the point: the command prints only its own echo, while the module loads, the breakpoint banner and anything an embedded `bp X "…; g"` script printed all arrive **during the wait**, and on the run that ends the target there is no successor to print them again.

**The session was then half-dead.** `.echo` and `.lastevent` kept answering while `k`, `r` and `registers` failed `0x80040205` — from a caller's side indistinguishable from #226's wedged session, which needs the opposite response.

**And execution control with no debuggee took the worker down.** A raw `g` there is a `STATUS_ACCESS_VIOLATION` *inside* DbgEng — a structured exception, so no `catch_unwind` traps it. Deterministic, not the intermittent race the issue describes, and it reproduces on a **fresh** engine that never had a target, which is what says the trigger is the missing debuggee rather than the departure.

Two corrections worth recording, since both were load-bearing:

- The issue reports the post-exit status as `0x7 DEBUG_STATUS_GO_NOT_HANDLED` and concludes the status can be stale. `0x7` is `DEBUG_STATUS_NO_DEBUGGEE`; `GO_NOT_HANDLED` is `3`. The status is reliable and is the **only** thing that still answers — `GetNumberProcesses`, `GetCurrentProcessSystemId` and `GetExitCode` all fail `E_UNEXPECTED`, and `.lastevent` says `<no event>`.
- Item 48 said the buffer holds `cmd.exe exited with code 0`. It does not — the engine will not say *how* it ended, only that nothing is left.

## What it became

Ending, not error: `target_gone` on the stop report and a sentence beside it, on **both** halves of the result, since a structured-aware client forwards `structuredContent` and drops the text. `run_to_address` gains a `target_gone` verdict, deliberately not a timeout: the run ended because the target did, so the address was never ruled out.

One refusal afterwards, not one per tool. dbgscope refuses raw commands itself, so without a gate here the typed tools would be the inconsistent half — `registers`, `modules` and `backtrace` reach the engine's own interfaces, which answer `E_UNEXPECTED`. `refuse_when_the_target_is_gone` runs once per op, exempting the openers (an engine before its target reads the same, and a worker is sent exactly one, as its first op), `end_session` (the answer the refusal gives) and `interrupt` (which never reaches the queue). Category `stale_session`, because no change to what is asked will help.

And a command can take the target away *itself*, which no pump ever sees: `.detach`, `q` and `qd` return with the engine already holding nothing. Read from the run rather than the command's name — `.kill` is measured **not** to be in that group, leaving a target that still reads a stack and going away on the *next* resume, so a name list would be wrong per engine version.

`engine_error`'s `_` arm was folding `StaleSession` into `Debugger` — the exact failure its own doc comment warns against — telling callers to change what they asked on a session where nothing they could ask would work. It had never mattered because nothing but the supervisor produced that category before. Caught by the test asserting the category rather than the message.

## Tests

Two in the launch tier, one per road to the ending: the typed resume, and the raw hatch's pump (the issue's minimal repro). Each asserts both halves, because either alone is satisfiable by something wrong — reporting the ending while leaving the chain in place fixes a message, and refusing everything afterwards without reporting it turns a program finishing into a call that failed.

dbgscope#120 carries three more, none `#[ignore]`d so CI runs them on both architectures, plus a **session fuzzer**: randomised command sequences checking after every step that the engine either still holds a target and answers or says it holds none. At `--seed 1` against pre-fix `main` it finds four violations in eight rounds; the same seed is clean on the fix, as are 150 rounds of 14 steps.

## Verification

ARM64 bench, debugger tier on: **577 unit + 87 smoke passed, 0 failed** (55s, so the tier ran). `cargo fmt --all --check`, clippy (one pre-existing `chunks_exact` warning in `src/dump.rs`), and markdownlint over the CI globs all clean.

The golden re-record is the whole cost of the two new fields: +49 bytes on each of the six stop-report tools, +40 on `run_to_address`, and **zero** `modelVisible` — output schemas never reach the model, so this is wire only.

## Left deliberately

`session_status` still reports such a session as `open`. Telling the supervisor needs a new `WorkerMessage` and a `SessionState` for it (`Retired` means "the worker still holds a target", which is the opposite), and the caller already has the fact — every call is refused with it and `end_session` works. Item 48 records it.


glslang:
Two follow-ups since the description, and one thing from the issue deliberately not done.

## The remote transport, measured rather than assumed

Both guards read `GetExecutionStatus`, and `enginehost`'s sessions live in another process — where `GetSymbolInformation` is already known not to cross, so "a core call obviously does" is not an argument. Getting this wrong would refuse every op on a 32-bit .NET dump, the feature that shipped this morning.

It crosses. Against a `cdb -server` on the ARM64 bench, a connected client reads `DEBUG_STATUS_BREAK` and runs commands normally. It *can* read no-debuggee — in the window where the host has published its pipe and not yet opened the dump, the race `await_hosted_target` was written for — and in that window `modules()` fails at the same instant:

```
on connect: status Ok("0x7"), modules Err(… 0x8000FFFF "Catastrophic failure")
modules after 2 polls: 227
execution_status -> Ok("0x6")
`lm` (guarded)   -> Ok(281)
```

Six runs, one of which caught the race, and the two never disagreed in the direction that would matter. That makes `await_hosted_target`'s "`modules()` is the cheapest question that cannot be answered until a target exists" load-bearing for something it was not written for, so both ends now say so (5d91773).

## `p` and `t`, and the commands that end a target outright

Steps need nothing extra — the same string, the same `execute_and_wait`, and the classification happens after the wait without reading the command. What did need something is the group the pump never sees: `.detach`, `q` and `qd` return with the engine already holding nothing, so `settle` finds nothing and the ending would have been reported by nobody. `raw_command` now reads it from the run.

`.kill` is measured **not** to be in that group — it leaves a target that still reads a stack in `ntdll!LdrShutdownProcess` and goes away on the *next* resume — which is the argument for asking the engine rather than listing command names, since such a list would need re-deriving per engine version.

One behaviour change worth naming: a `debug_batch` whose `always` block runs steps *after* a `.detach` now gets a clean `NoDebuggee` per step instead of the `0x80040205` (or, for anything reaching execution, the access violation) it got before. Strictly better, and no test asserted the old shape.

## Not done: structured narration of script errors

The issue's optional third ask — "engine script-error narration could be structured (e.g. `script segment N failed`) so a raw HRESULT inside command output is not mistaken for a server failure" — is declined. That would mean parsing DbgEng's own output text to find which segment failed, which is the thing this codebase refuses to do on principle ([#77](https://github.com/glslang/windbg-mcp/issues/77): a figure recovered from a rendering measures the rendering). The issue itself already says the `0x8000FFFF` a second resume produces is DbgEng correctly refusing an illegal operation and should stay; what made that confusable with a server failure was the *session* being half-dead afterwards, and that is what this fixes.

glslang:
Codex's finding is right, and it is the same defect one layer up. Verified against `batch::run` and `probe_state` before acting on it: with `execute_and_wait` now returning `Ok` for a target that ran to completion, a `debug_batch` read an ordinary success, ran every later step against a session with nothing to run them on, and — where the resume was the last step — reported `COMMITTED` with `after: uncertain`. That `uncertain` was `? @$ip` failing with *"No active debuggee"*, which is the batch being told the answer and reporting that it did not know.

Fixed in d627e5e. The fact travels as a value the way `cut_short` does, and for the same reason: a step that ended the target *succeeded*, so every reader that had to infer it from prose would sooner or later not. `Ran::target_gone` → `StepOutcome::target_gone` → `BatchOutcome::TargetGone { at }`, later steps skipped, `SessionAfter::Ended { by }` naming the step.

Three decisions inside it worth stating, since each could reasonably have gone the other way:

- **`Ended` is not `Detached`.** Both mean "this session has no target", and both now reach the batch as `target_gone`. But a detached process is still running somewhere and can be attached to again, and on a live kernel that is the difference between a machine that is up and one that is not. `probe_state` keeps its release check first, so `.detach` still says detached; what reaches the new arm is the ending no command name announces. Pinned both ways.
- **Its own outcome, not `Failed`.** Nothing failed. And not `Interrupted`, whose whole advice — "resubmit on this session" — is the one thing that cannot work here.
- **The `always` block is still attempted, not skipped with a reason.** Skipping reads tidier and is the wrong direction: if the ending were ever misread, attempting cleanup costs one refusal per step, while skipping drops cleanup that could have run. The refusals are also the proof of *which* cleanup did not happen.

ARM64 bench, debugger tier on: 579 unit + 87 smoke passed, 0 failed. Golden re-recorded: +213 bytes on `debug_batch`'s output schema, zero `modelVisible`.

glslang:
The two later findings are both right, and both are the same shape as the bug itself: a fact that exists on one channel and not the next. Fixed in 32327ac and 5a25a29.

**The transcript.** `record::derived` copied `stopped_at`, `interrupted` and `timed_out` into `Event::Stop` and dropped `target_gone`. An ending carries no position — like a module-load break — with both other flags false, like an ordinary stop, so a reader saw a locationless stop followed by every later call on that session being refused, with nothing joining the two. `--render-cast` had the same gap and printed *"`g` stopped at an unknown position"*; it now says the session is over. The test that already existed for this exact class — its own doc comment calls it "a fact that exists in one channel and not another" — now covers three reasons rather than two.

**The description.** Three doc comments enumerate three verdicts and there are now four: the tool's own, `debug_batch`'s `run_to` step, and the batch action. An output schema never reaches the model, so the description is the only place a model could learn there is a fourth, and one that lists three when there are four is *wrong* rather than merely incomplete — which is the bar for changing prose here.

`go`'s description is deliberately **not** touched, and that is the same rule applied the other way. "Runs to the next breakpoint, or the end of a TTD trace" was never a closed list — an exception, a module load and the watchdog's own break were all already missing from it — so adding to it would cost model-visible bytes to correct nothing.

Naming the fourth verdict costs **107 model-visible bytes**: 67,766 → 67,873, `exec` 3,420 → 3,475, `debug_batch` 9,746 → 9,798. That number is quoted as a live fact in four places (`README.md`, `CLAUDE.md`, `toolset.rs`'s module doc, a smoke-test comment) and all four are re-derived from the golden. `docs/token-budget.md` and `docs/local-model.md` are left alone on purpose: the first says in its own header that the golden is the authority and its tables are "a reading of it at the time of writing", and the second records what an eval run was measured against — editing either would falsify a dated measurement rather than correct a stale claim.

ARM64 bench, debugger tier on: 579 unit + 87 smoke passed, 0 failed.

glslang:
All three of the latest round hold, each checked against the code first, and all three are the same shape as the bug they follow: a terminal fact that reaches one caller and not the next. Fixed in 3f7cd30.

**The raw hatch's pump did not put the ending on its run.** `raw_command` read `settled.target_gone` only to pick the prose; the `CommandRun` it returned still said `false`, because the *command* had not ended the target — the pump had. So `debug_batch` with `{"op": "command", "command": "g"}` got an ordinary success and ran on. Worth saying explicitly that this is **not** a repeat of the earlier finding: different function, different route to the same state, and the fix for `resume` genuinely did not cover it.

**Assertions were sent to an engine with no target.** An `eval` check is engine calls, and a refused `? (...)` reads back as `Unmet` — a false failed assertion on a step whose action did exactly what it was asked. Engine-backed checks now stop at the ending and the step says why: `Failed`, not `Unmet`, for the reason the deadline case beside it already gives in as many words — nothing was learned about the target. `Check::needs_the_engine` is a `match` rather than a negation, so a check added later is a compile error here instead of defaulting to the safe-looking wrong answer.

**A cleanup that could not run took credit for the ending.** `probe_state` scanned for a release before an ending, and `changes` is recorded whether or not a step succeeded — deliberately, since a command that errors may already have written. So a `.detach` in the `always` block, refused because the target was already gone, was reported as having detached it. Now one ordered scan: whichever step took the target away *first* did it, and how — released or ended — is read from that step.

One thing worth recording against the obvious simplification: the `changes` test **cannot** be dropped in favour of the typed one. `.opendump` *replaces* the target, so the engine still has a debuggee and `target_gone` is false, while the handle is just as retired. Both tests are load-bearing and for different commands.

Three tests, one per finding. ARM64 bench: 581 unit + 87 smoke passed, 0 failed.

glslang:
Right again, and this one is self-inflicted by the previous commit. Fixed in 16793d2.

`BATCH: TARGET GONE` said *"Nothing failed: the step did what it was asked"* unconditionally, while the same step could be sitting in the list below as FAILED — either because a `contains` expectation on its output did not hold, or because its `eval` expectation could not be checked at all once there was nothing left to ask, which is exactly what the last commit introduced.

Both facts are true of one step and only one of them fits in the outcome. The outcome stays `TargetGone`, because that is what decides the caller's next move — open a new session, and nothing they ask this one will work — and `committed` is false either way. The *sentence* is now conditional, so the headline stops contradicting its own report, and the step list says which expectation it was.

Both halves pinned, since a conditional that is always one way is the same bug with more code.

582 unit + 87 smoke passed on the ARM64 bench.

glslang:
Right, and this is the last route the ending had left. Fixed in ba0b1d1.

`raw_command`'s own contract is that a command failing part-way may already have run an earlier segment that resumed the target — `g; bogus` is the shape — and the pump that follows is where the process exits. The failure comes back as an error *string* with the ending only in appended prose, so nothing the step returns can carry it.

Unread, that was not merely vaguer. `probe_state` reached its *"a step resumed the target and did not report a stop"* arm and reported the session **RUNNING** — the most wrong of the answers available, on a session with no target at all.

So the engine is asked, through a `Debuggee::has_target` beside the `running()` already there. That closes the class rather than this route: any path where a step's own answer cannot carry the fact now has somewhere to get it.

Two things fell out of it, both worth naming:

- The scripted debuggee marks the ending from the **call** rather than from its result. A command that fails can still end the target, and a script that only marked successful calls could not express the case under test — which is how this survived the previous round's tests.
- The target-gone headline's conditional clause now covers an action that failed, not only an expectation that did not hold. The previous commit's wording assumed the only way such a step could be not-`Ok` was its assertions, and this makes that false.

583 unit + 87 smoke passed on the ARM64 bench.

glslang:
This one I am **declining here**, with the fact taken — recorded in `FOLLOWUPS.md` item 48 in b39fff2, which is where it already lived as a deferral.

The finding supplies a consequence my own deferral note did not have, and it is a better reason than the one I wrote: a dead session is still the **default route**. `Registry::current` takes the most recent session whose state `accepts_default`, and `Open` does — so every call naming no `session_id` goes to the session with no target, and a healthy older session of the same client is shadowed by it. That is more than the wrong status string item 48 claimed.

What decides the remedy is the other half, which I checked before answering: **it is pre-existing.** On `main` the only state transitions are `Closed` (teardown, worker death, the sweep), `Failed` (an open that created nothing), `Retired` (a target-changing *command*, decided from its text) and the opening pair. Nothing has ever watched a target leave, so a session whose process exited was already `Open` and already the default route. What this PR changes is that it now says so on every call instead of failing with `0x80040205`.

So: real, worth fixing, and not this PR's. It is session lifecycle — the part of this server where a mistake costs a *target* rather than a call, a live kernel left halted — and it is orthogonal to the ending. It wants its own change, its own tier run, and a reviewer looking at nothing else.

Item 48 now carries the design rather than the symptom: the milestone to emit (the worker already asks `has_target` once per op in `refuse_when_the_target_is_gone`, so there is one place), a `SessionState::Ended` that refuses `accepts_handle` *and* `accepts_default` while staying `is_live` — the worker exists, owes an `end_session` and counts against the four-session cap until it gets one — why `Retired` cannot be reused, and the opener-settle promotion rule that would have to protect it the same way it already protects `Retired`.

glslang:
Right, and the fix is exact. d1594d3.

The `always` block is still attempted after an ending — deliberately, so a misread ending cannot drop cleanup that could have run — and each of its steps is then refused by an engine with no target. Asking the engine *afterwards* cannot tell "it went during this step" from "it was gone before this step", so every refused cleanup step reported itself as the ending, and a consumer looking for the step that did it found several.

So the answer is carried forward rather than re-derived: one `ended` bool threaded through both blocks, and a step is only credited with an ending the target had not already had. The `Ok` branch is gated by it too, though nothing can reach that — dbgscope refuses a command outright when the engine holds no debuggee, so a *successful* call cannot have run against an absent target. One rule with no exemptions is easier to keep true than one rule plus an argument about why the other branch is exempt.

Worth being precise about the blast radius: **behaviour was already right.** `probe_state` takes the first such step and the main loop never sees the `always` block, so `after` and the outcome both named the real one. What was wrong is the *field*, which documents itself as "this step ended the target" and was true of steps that did not — a structured-consumer defect rather than a batch one, which is exactly where a typed report has to be right.

The invariant is now pinned rather than left implicit: at most one step in a report carries it, asserted over both blocks together.

584 unit + 87 smoke passed on the ARM64 bench.

glslang:
CI went red on `schema::tests::the_same_type_is_walked_once` — on an x64 runner, on a commit that touches only `src/batch.rs`, and green on the ARM64 bench and on every other run of the same code. A flake, with a real race underneath. Fixed in bc832ee.

`constraints_of` drops the read lock **before** walking the type, deliberately, so a walk cannot block every other type's lookup. Two callers can therefore miss together and both arrive at the write with a document of their own, and a plain `insert` let the second overwrite the first. One type then has two `Arc`s alive, and a caller that asks twice can be handed two different ones — which is exactly what that test asserts against, and what it caught.

The window needed only the scheduling: three tests in `schema.rs` ask for `OpenOutcome`, and every test that builds a `router()` asks for it again. This PR added tests, which is presumably what moved the interleaving; `src/schema.rs` is untouched by it (empty diff against `main`), so it did not create the race.

`or_insert` keeps whichever landed first and hands it back, at the cost of the loser's wasted walk — once per type at most. **Fixed in `constraints_of` rather than by serialising the test**, because a test made serial would have hidden a cache that really can hand one type two documents.

Six consecutive runs of the unit suite on the bench: 584 passed, 0 failed, each time. That is weak evidence against a race by itself — the honest claim is that the interleaving is now unreachable by construction, not that it was run out of existence.

PR #252, opened 2026-08-28
docs: the release procedure was three files short and lived in my head

glslang:
`docs/releasing.md` described cutting a release as *"bump `plugin.json` and `Cargo.toml`, bump the
badge, add a CHANGELOG entry, and tag"*. Every clause of that is true, and it is not enough to cut
one with. The rest lived in an assistant memory — the wrong place for it twice over: nobody else can
read it, and it drifts against `release.yml` with nothing to catch the drift.

## What the old sentence left out

| | Why it bites |
| --- | --- |
| **`Cargo.lock` is a fourth file** | It holds this crate's own version, and `release.yml` builds `--locked`. A stale lock fails the build *after* both version guards have passed — the expensive place to find out |
| **Keep an empty `## [Unreleased]`** | *Renaming* it to `## [X.Y.Z]` orphans the `[Unreleased]:` link definition and **Documentation lint fails on MD053**. That is the one gate that catches release prep, and it catches it after the push |
| **Wait for `main`'s CI before tagging** | The tag push publishes to the MCP Registry, where a version number is spent. A failed build re-runs; a published `io.github.glslang/windbg-mcp` version cannot be taken back |

## What else it now records

- **The flow.** Release prep goes straight to `main` on the admin ruleset bypass — sanctioned for
  that and nothing else; every other change takes a PR. Read the ruleset with
  `gh api repos/glslang/windbg-mcp/rules/branches/main`: the branch-protection endpoint 404s here,
  because protection is a *ruleset*, which reads as "no protection" if you stop there.
- **Dry-run the guards** with the workflow's own regexes rather than by eye — twenty lines into
  `release.yml`, a second to run in `pwsh` — and the lint with CI's own globs.

## One claim corrected by measuring it

The old text put `claude plugin validate . --strict` next to the version bumps, which reads as a
check on them. Run in this repo it validates `.claude-plugin/marketplace.json` and names only that
in its output, and it passes identically before and after a bump — it never sees a tag. The version
agreement is `release.yml`'s guard and nothing else's, and the doc now says which is which. Still
worth running; it just answers a different question than its placement implied.

## Provenance

Everything here was executed in cutting **0.13.1** (`97c1078`) rather than recalled: the four bumps,
the CHANGELOG link-definition edit, both guard dry-runs, and the lint. The bypass rule and its
release-only scope are the user's own, confirmed 2026-08-08 and 2026-08-15.

## Verification

`markdownlint-cli2@0.23.2` over CI's globs: **0 issues in 0 files**, 36 files linted. Documentation
only — no code, no manifest, no workflow touched, so nothing here affects the 0.13.1 release already
in flight.

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -54,10 +54,29 @@ protection" if you stop there.
    file unreferenced, and **Documentation lint fails on MD053** — the one gate that catches release
    prep, and it catches it after the push. Add the new `[X.Y.Z]:` definition beside it and repoint
    `[Unreleased]:` at the new tag. That section *is* the GitHub release body.
-3. **Dry-run both guards locally**, with the workflow's own regexes rather than by eye — they are
-   twenty lines into [`release.yml`](../.github/workflows/release.yml) and take a second to run in
-   `pwsh`. One compares the tag against all three manifest versions, the other requires a
-   `## [X.Y.Z]` heading in the CHANGELOG. Dry-run the lint too:
+3. **Dry-run both guards locally**, rather than by eye. They are twenty lines into
+   [`release.yml`](../.github/workflows/release.yml) and take a second in `pwsh` — but copy them
+   with one change, or the dry-run is worse than not running it. The workflow reads the tag from
+   `$env:GITHUB_REF_NAME`, which Actions supplies and your shell does not, so a **verbatim** copy
+   compares all three manifests against an empty string and reports a mismatch on a correct tree.
+   Set `$tag` yourself. The second guard needs no tag at all: like the workflow, it takes the
+   version from `Cargo.toml`, which the first guard has just pinned to the tag.
+
+   ```pwsh
+   $tag    = '0.13.1'   # the version about to be released, without the leading v
+   $cargo  = [regex]::Match((Get-Content Cargo.toml -Raw), '(?ms)^\[package\].*?^version\s*=\s*"([^"]+)"').Groups[1].Value
+   $plugin = (Get-Content .claude-plugin/plugin.json -Raw | ConvertFrom-Json).version
+   $readme = [regex]::Match((Get-Content README.md -Raw), 'img\.shields\.io/badge/release-v([^-]+)-').Groups[1].Value
+   if ($cargo -ne $tag -or $plugin -ne $tag -or $readme -ne $tag) {
+     throw "Version mismatch: tag v$tag, Cargo.toml $cargo, plugin.json $plugin, README badge $readme"
+   }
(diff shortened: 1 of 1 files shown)


glslang:
Fixed in 31d480d — the finding was right, and it is the sharpest kind: a step that fails on a correct tree.

The first guard derives `$tag` from `$env:GITHUB_REF_NAME`, which Actions supplies and a local shell does not, so a verbatim copy compares all three manifests against an empty string and reports a mismatch on a tree where nothing is wrong. That is worse than skipping the step — it teaches the operator to disbelieve the guard they are about to be gated by. Step 3 now carries the self-contained block with `$tag` set by hand, and says why it differs from the workflow's copy.

One detail worth adding to the finding: only the *first* guard has this problem. The second takes its version from `Cargo.toml`, exactly as the workflow does (it is ungated there so it can run on `workflow_dispatch`, which has no tag), so it needed no change. The doc now says so, which is what stops the next reader wiring up an env var for both.

Verified rather than read: the block runs verbatim against the branch and prints `guards pass: 0.13.1`; with the tag changed to `0.13.2` it throws `Version mismatch: tag v0.13.2, Cargo.toml 0.13.1, …`, so it catches what it exists to catch; and `GITHUB_REF_NAME` is confirmed empty in this shell.

glslang:
Fixed in c765555 — reproduced with an empty `CARGO_HOME`: exit 101, `can't checkout from 'https://github.com/glslang/dbgscope': you are in the offline mode (--offline)`, before `Cargo.lock` is touched.

**One of the two suggested remedies does not work here**, which is worth recording so nobody takes it later. "Add an explicit fetch/bootstrap step" would naturally be `cargo fetch --locked` — this repo already recommends that elsewhere for populating the cache without compiling. At this point in the checklist `Cargo.toml` is *ahead* of `Cargo.lock` by construction, since bumping the lock is the step, so `--locked` refuses: `cannot update the lock file … because --locked was passed`. Measured.

So the flag goes, and what replaces it is a fact rather than another flag: **`-p` is what confines the update**, which is what I had wrongly attributed to `--offline`. Measured both ways — warm cache, the bare command reports `42 unchanged dependencies behind latest` and leaves the lock byte-identical to what was committed; cold `CARGO_HOME`, it fetches the crates.io index and the `dbgscope` git repository, reports 45, and moves the one line. The doc now says that, and names `git diff Cargo.lock` as the check.

Both rounds on this PR have been the same shape: a command I ran successfully on a warm, already-configured bench, written up as if the state it depended on were part of the instruction.

PR #257, opened 2026-08-29
Asynchronous execution handles: continue_async, wait_for_stop, break_in

### LINKED ISSUE #83: Add asynchronous execution-control handles
## Problem

`go`, stepping, and `run_to_address` synchronously wait for the next stop. The MessageManager workflow needed WinRM, separate processes, guest-side delays, and manual handoffs to launch work while KD was running.

This makes a common debugger sequence awkward:

1. arm a breakpoint;
2. resume the target;
3. start or signal work outside the debugger;
4. wait for the breakpoint;
5. interrupt or cancel safely if it does not arrive.

Guest `Sleep` is not a sound synchronization primitive while KD has halted the guest clock, and a second ad-hoc MCP client should not be required.

## Proposed shape

Add session-scoped execution primitives such as:

- `continue_async` → returns an execution handle after the target is resumed;
- `wait_for_stop` → waits on that handle and returns the stop reason;
- `break_in` → requests an interrupt for the active execution;
- optional `execution_status`/cancellation.

Only one execution handle may own a session at a time. Ordinary target-reading tools should give a clear “target is running” response instead of queueing ambiguously behind it.

## Acceptance criteria

- A caller can resume, perform unrelated host/guest orchestration, and then wait for the stop.
- Stop reason, thread, processor, and instruction pointer are returned explicitly.
- Timeout/cancellation does not abandon a hidden `WaitForEvent`.
- `end_session` can safely break in, resume as appropriate, and detach while an execution handle exists.
- Client disconnect has a documented and tested target-release policy.
- Tests cover handle ownership, duplicate continue, wait timeout, break-in, and teardown.

glslang:
Closes #83.

`go` waits for the next stop and answers with it, which leaves nowhere to put the step every live-target sequence needs: **arm a breakpoint, resume, make the thing happen that trips it, then collect the stop.** The MessageManager work routed around that with WinRM, guest-side delays and a second ad-hoc MCP client — and a guest `Sleep` is not a synchronisation primitive while KD holds the guest clock.

## The design decision worth reviewing first

**The wait that goes away is the caller's, not DbgEng's.** A target moves only while the engine thread is inside `WaitForEvent` — `Execute` sets the run state and returns without the target having moved, which is the fact [#226](https://github.com/glslang/windbg-mcp/issues/226) turned on. So there is no arrangement in which the worker starts a run, returns to its queue, and the target keeps going: it would sit still. `EngineOp::Resume` changes only *when the reply crosses the pipe* — the engine thread executes, reports a new `WorkerMessage::Resumed` milestone, and stays in `settle` for the whole run.

Which makes the **supervisor** the owner of the run, and everything else follows from that:

- The execution slot is filled **before the job is queued**, so there is no window where the target may be moving and this side would let a read through.
- The reply is filed by a task spawned **before the call waits for anything**, because the case it exists for is the caller not being there: a client that disconnects mid-run must still leave a stop somebody can read.
- Filing is keyed by **job, not handle**. The task outlives its caller, so a reply matched by handle alone could land the run-before-last's stop on the run in flight — reported as a target that had stopped when it is still moving, which is the one answer here that gets somebody to read a moving target.

## Acceptance criteria

| From the issue | Where |
|---|---|
| Resume, orchestrate, then wait | `continue_async` → `wait_for_stop`, `docs/sessions.md` |
| Stop reason, thread, processor, IP returned explicitly | `StopReport` — `stopped_at`, `thread`, `processor`, plus `interrupted` / `timed_out` / `target_gone` |
| Timeout/cancellation does not abandon a hidden `WaitForEvent` | the run is recorded before the job is queued, `session_status` reports it, `max_run_ms` bounds the pump, and `wait_for_stop` waits on a channel rather than the engine |
| One handle per session; other tools say "target is running" | `Session::claim_execution`; `EngineError::TargetRunning` / `ErrorCategory::TargetRunning`, refused rather than queued |
| `end_session` can break in and detach while a handle exists | `worker::break_any_pump`, from the request reader beside `announce_teardown` |
| Client disconnect has a documented and tested policy | it runs the same release, so `a_session_can_be_ended_while_its_target_is_running` covers it; `docs/sessions.md` states it |
| Tests: ownership, duplicate continue, wait timeout, break-in, teardown | 11 unit tests + 2 debugger-tier tests |

Two refusals are worth calling out because they are decisions rather than mechanics. **Work on a moving target is refused, not queued** — queued, a `registers` would be answered whenever the target next stopped, up to an hour away, about wherever the target happened to be, and from the caller's side that is indistinguishable from a hung debugger. **A wait that runs out is a poll, not a failure** — no stop, nothing cancelled, handle still good — because a caller reading it as a failure would end a session that is working perfectly.

## Depends on glslang/dbgscope#123

**Merged.** It was rebase-merged, so the branch head this pinned was an ancestor of nothing on `main` — `merge-base --is-ancestor` said so, which is why that check is the rule rather than assuming a merge commit exists. `Cargo.toml` now pins `8474b56`, which is `dbgscope`'s `main`.

## CI result

`620 passed; 0 failed` unit tests (613 + the seven added here) and `91 passed; 2 failed` in `mcp_smoke` on all four jobs of run [33296992550](https://github.com/glslang/windbg-mcp/actions/runs/33296992550), the two being the goldens — since re-recorded. On the ARM64 bench with the goldens in place and the debugger tier on, `mcp_smoke` is **93 passed; 0 failed; 12 ignored**, with the only `SKIPPED` lines being the TTD tier CI also skips.

**Both new debugger-tier tests pass against a real DbgEng on a launched target**, on all three architectures — the whole lifecycle: the handle issued while the target is moving, a second resume refused, a read refused as `target_running`, a short wait polling without disturbing the run, `break_in` producing an `interrupted` stop with `stopped_at` and `thread` present and `processor` absent, the stop read twice, and `end_session` releasing a target that is still running.

CI runs that tier on every PR, on three matrix entries. An earlier version of this description said the tests had never run; that was wrong.

## The number worth reviewing

**These three tools cost 4,892 B of model context** — `continue_async` 1,987, `wait_for_stop` 1,530, `break_in` 1,375 — which is 7% of the surface, paid by every client served `exec` at the start of every conversation. `MODEL_VISIBLE_CEILING` goes 76,000 → 83,000 to keep the +12% headroom its own comment specifies; the ceiling is bookkeeping, the 4,892 is the decision. `--tools` is the answer for a client that cannot afford it.

The goldens are re-recorded and the surface is **73,996 B across 54 tools**, from 68,893 across 51. The comment's arithmetic predicted 73,785, and the 211 B it missed is the interesting part: `interrupt` grew without being touched, by a `TOOL_NOTES` cross-reference to `continue_async` that is appended only where that tool is served. The sum of the new tools was right and the model of the surface was not — which is the failure mode a derived figure has and a recorded one does not.

That same mechanism is why two of the four `--tools` rows did not move at all: `session,inspect,crash` and `crash` do not serve `exec`, so the note is not there, and both are byte-identical at 25,465 and 14,587. Those rows are measured, not assumed — each is a `tools/list` off a server started with that spec, by a method that reproduces the golden's own totals exactly (73,996 and 192,936).

Everything downstream moved with them: `docs/tool-surface.md` (`session,inspect,exec,crash` is 28 tools and 33,985 B), `src/toolset.rs`'s group table, `docs/token-budget.md`'s two tables and share percentages, `README.md`, `docs/remote-listener.md`, `docs/local-model.md` and `CLAUDE.md`. Finding 8's prose split was re-derived too and had gone stale in a way worth naming — 74% / 24,902 / 25,333 was a percentage of **67,873**, a surface total two re-derivations out of date; its own arithmetic was never wrong. It is 70% of 73,996 now, with the method written into the file. The `51`s in `docs/local-model-eval.md`, `README.md`'s eval line and `FOLLOWUPS.md` item 24 are left alone: they record what a published run measured.

### One thing left for a human

**Four CodeQL `rust/cleartext-logging` alerts**, all "writes `started.session_id` to a log file" against `describe_started` in `src/server.rs`. They are not log statements — that function builds the tool's *result text*, returned to the caller who supplied the session id. Alerts #3 and #4 are the same rule, dismissed as false positives on 2026-08-28 with a rationale that applies verbatim: a session id is a routing handle, not a credential, and `resolve()` reports one whose owner is not the caller as unknown, so possession grants nothing. Dismissing a security alert is not mine to do.

## What the first CI run caught

Worth recording because the bug was mine and the tier is why it is not shipping:

- **`continue_async` waited for the stop it exists not to wait for.** `start_execution` waits on the execution slot, which nothing touches until the run *ends*, so the milestone woke nothing: the call blocked for exactly as long as the target ran, then reported `running: true` about a run that was over. Introduced in this branch's second draft, when a `select!` over the milestone and the slot became "read the milestone, then wait on the slot". Every test written for the ordering passed either way. Fixed with one wait — `reader` bumps the slot when the milestone lands — and `a_run_is_handed_back_while_the_target_is_still_moving` is the assertion that was missing: it sends the milestone and **no** reply, so a call that waits on the stop cannot pass it.
- Five hardcoded `51`s in `src/client.rs` and `mcp_smoke` (counts `Toolset::summary` renders, not measurements).
- The output-schema coverage gate, which requires every schema-bearing tool to be exercised or excused.
- My smoke tests read `category`/`message` at the top of the error payload; they are under `error`.

## Review: twelve findings, and most of them were two choices

Fixing each in place would have been a mechanism apiece. Two seams generated the bulk of them, and the seams go instead.

**A break was session-wide and its effect was inferred** (four findings). `break_in` sent the same unqualified `EngineOp::Interrupt` the `interrupt` tool does, then read `Ok` as delivery. It now carries the job it is for: `None` stays the `interrupt` tool, whose caller named a *session*; `break_in` holds a handle to **one run**, and between reading the slot and the request being read the run can stop and the engine thread start the next thing — a queued command, or the run after this one — which an unbound break would cut short and report to *its* caller as an interruption nobody asked for. Named, the worker refuses rather than rebinds.

That also makes delivery answerable. Four of the five things an interrupt can do are `Ok` and raise nothing, so the reply carries `Output::raised` and `requested` is the worker's answer rather than `is_ok()`. `false` now has one meaning — the target had already stopped — and a break that could not be **delivered** is an error, where a lost worker used to read exactly like that benign race. The transcript's `interrupt` event takes `delivered` from the same answer: it exists to explain a *later* truncated result, and one recorded against a break that was never raised attributes a short answer to a request that did nothing.

**A teardown could only reach a run that had already started** (two findings). Breaking the pump finds nothing when the resume is still queued — behind a long `pool_census`, or simply not yet dequeued — so the teardown queued behind it and waited out the whole run it had arrived too early to stop: the grace expiring against a worker that was working perfectly, and the worker killed still holding the target. `break_any_pump` becomes `stop_resuming`, which also sets `Running::tearing_down`; `resume` claims the pump through `claim_pump`, which refuses while that is set. Read and written under one lock in opposite orders, so whichever gets there first the other sees it. The EOF path calls it too, where the reason is sharper: that release waits `ABRUPT_EXIT_RELEASE` — five seconds — against a run that may have an hour left.

Four smaller ones, each a line:

- **`Session::submit_gate`.** Reading the execution slot and enqueueing are two steps; between them an ordinary call could pass the check, be descheduled, and enqueue *behind* a resume that claimed the slot — waiting out the run instead of being refused, and if the run outlasted its call timeout its caller saw a timeout while a mutating command still ran when the target stopped. A lock of its own rather than the slot's, because the enqueue takes the registry lock while `snapshot` takes the registry lock and then the slot's.
- **`Session::busy` reads the slot too.** `reader` removes the waiter when the reply lands and the filing task runs after it, so in between a session with a stale submission stamp looked idle — and a concurrent open could close it, leaving the stop filed where nothing could resolve it.
- **`Execution::ran_for` is stamped at the stop.** Derived from `started.elapsed()`, a stop read an hour later reported an hour, and two reads of one stop disagreed.
- **`wait_for_stop` takes one deadline.** The slot is bumped by things that are not stops, so re-arming per wakeup made `timeout_ms` a per-wakeup allowance — `STOP_WAIT_MARGIN` not holding.

**Two declined.** Tracking a queued run's phase apart from a moving one: the slot is filled before the job is queued deliberately — filled when the worker answered, it would leave a window where the target may be moving and this side would let a `registers` through — so reporting it as running is the conservative answer, and the only thing that misread it was the unqualified break above. And four CodeQL "cleartext logging" alerts on `started.session_id`: a session id is this server's own handle, minted here and logged on every other path already.

Seven new tests come with it, five in `engine` and two in `worker`; the two worker ones stage both orderings of the teardown race.

## Review rounds on the dbgscope side

Four rounds, and the facts landed in opposite places — two moved the code, one moved a sentence.

- **`.thread` is not a case where no processor applies** (raised twice) — it changes the register context and leaves the current engine thread as the processor's. Fact taken, remedy declined: inspecting the implicit-thread context answers a different and worse question, since on a kernel target the CPU really is where the break is, and reaching the selected thread's id means reading `ETHREAD` through symbols rather than asking the engine. Both doc comments grew the qualifier they were missing; the code did not change.
- **A failed lookup is not a processor that does not match.** `is_ok_and` folded a failed `GetThreadIdByProcessor` into "no match", so a dropped KD link could answer `Ok(None)` — absence, where the truth is unknown, which is the shape `docs/unknown-not-absent.md` exists to refuse. Now three cases: a match wins whatever else failed, no match with every lookup answered is a real `None`, no match with a failure among them is `Err`. `windbg-mcp` collapses both into an absent field, deliberately and now explicitly — a model cannot act on the difference.
- **A repeat of a lookup that answered can only make things worse.** The candidate tried first was left in the scan behind it, so it was asked twice — and the second ask can only *fail*, which under the rule above would overturn a `None` every distinct processor had agreed on.

## One thing found and not fixed — `FOLLOWUPS.md` item 52

`no_description_names_a_tool_the_client_cannot_call` walks tool **descriptions** and not **input schemas**, and an argument's doc comment lands in the schema, which is model-visible. There is a live pre-existing instance: `RunToAddressArgs::address` says "Typically a block from `reachable_from_dispatch`", and `run_to_address` is `exec` while `reachable_from_dispatch` is `ioctl` — so any surface with the first and not the second ships a pointer its client cannot follow. These tools' argument docs were reworded to name no tool rather than adding a second instance while reporting the first; the item says what closing it would take.

https://claude.ai/code/session_01CZu3mUSEzwtjPod3n88vhD


glslang:
@codex review

PR #286, opened 2026-09-05
feat: exception_triage and decode_error_reporting — the user-mode fault, as fields

glslang:
Closes the gap [`docs/explorer-crash-walkthrough.md`](https://github.com/glslang/windbg-mcp/blob/main/docs/explorer-crash-walkthrough.md) §9 argued for. Three faults in one evening whose answer came out of `execute` — `.exr`, `.ecxr`, `s -d`, `dd`, `!error` — are one call now. Built on [dbgscope#144](https://github.com/glslang/dbgscope/pull/144), which merged first.

Against the checked-in fixture:

```text
EXCEPTION: 0xc0000409 at 0x00007ff79e442989
  The system detected an overrun of a stack-based buffer in this application. […]
  second chance — nothing in the target handled it, noncontinuable
  Parameter[0]: 0x0000000000000007
WHAT THIS IS: a __fastfail — a deliberate process exit, not a stack buffer overrun, whatever the
  code's name and the system's message text for it say. FAST_FAIL_FATAL_APP_EXIT (subcode 0x7) is
  what says why; subcode 7 is the CRT's abort(), which means a C++ exception nobody caught.
THROWN: hresult_error at 0x00000054f7cff950
  carries 0x80670015 — The StateRepository cache is not initialized.
```

The second and third lines together are the point: the misleading system message is **reported**, because a reader who has met it in an event log needs to place it, and then corrected.

## What lands

- **`exception_triage`** — the exception record with its code decoded, what kind of fault it is, the thrown C++ object and the HRESULT it carries, and the crashing stack. The stack comes from the **stored crash context**, so it is the crash whatever thread the session has selected, and the selection is left where it was — which is why this is annotated read-only where `crash_triage` needed a scope guard.
- **`decode_error_reporting`** — `!error` as a **pure host call**, not a scrape of an extension's output. `FormatMessageW` was measured to answer verbatim for both exotic codes from that investigation (`0x80670015`, `0x80073D54`), with `ntdll`'s table beside it for an `NTSTATUS`. It reports every *reading* rather than choosing one, and takes a sign-extended 64-bit value because that is the shape a WIL fail-fast's HRESULT arrives in.

Three things differ from the walkthrough's recipe, each a correction rather than a refinement. The thrown **type** is decoded from MSVC's `ThrowInfo` → `CatchableTypeArray` → `CatchableType` → `TypeDescriptor` chain, so the `0xAABBCCDD` sentinel confirms an offset the type name already predicted instead of being what finds it. The throw's own record is **searched for**, since the event the debugger sees is `abort`'s fail-fast. And both routes run, because neither is a superset.

## The fixture, and the bug it found

`docs/samples/cppthrow-fastfail.dmp` (175 KB) is the first user-mode sample here and the only purpose-built one — the faults that motivated this could not be checked in at 122–345 MB of a live desktop. `docs/samples/cppthrow.cpp` is the whole program.

**It immediately found a bug no synthetic test could.** On x64 the unwind data lives in the image's `.pdata`, which a minidump does not capture — so on any host without the fixture's own binary (every host but this one, CI included) the engine unwinds badly and the stack comes back with holes. The throw scan was bounded by `min(stack_offset)..max(stack_offset)` of that walk, so it scanned the wrong memory and **found nothing on exactly the dumps it exists for**. It is anchored on the innermost frame and a fixed span now: that stack pointer comes from the recorded context rather than from unwinding.

That is what makes "two routes, neither a superset" true rather than merely claimed. `docs/limitations.md` carries the unwind caveat, because it costs a reader a stack rather than a footnote.

**And the environment block is the thing to know about making such a fixture.** The first attempt carried this machine's user name, its whole `Path` and the harness's own variables — a minidump captures the PEB, and none of it shows in anything a debugger prints. It is regenerated with a scrubbed environment, and `docs/smoke-test.md` says to check before committing.

## Verification

- `cargo test`: **658** unit (was 656) and **100** smoke (was 99); `cargo clippy --all-targets` clean; `cargo fmt --all --check` clean; markdownlint clean over its four globs.
- The debugger tier was **run**, not assumed: `a_user_mode_fault_is_triaged_from_its_exception_record` passes with the fixture's image present *and* absent — checked by renaming the binary aside, which is the branch this bench is not on. It prints a `SKIPPED` line in the second case naming the claim it could not check.
- Both tools driven end to end against a real dump over stdio, not only through the harness.
- The 32-bit worker was rebuilt and restaged after every edit, per `.claude/rules/worker-architecture.md`.

## Surface cost

56 tools and 80,177 B of model context, from 54 and 75,547 (measured 2026-09-05). `--tools crash` is 13 tools and 18,378 B. The wire ceiling goes 205,000 → 215,000, and the arithmetic is what says it is headroom rather than a leak: the payload went 199,058 → 208,456, and the only row that moved in the review round was `exception_triage`'s own input schema, by exactly the 352 B of `scan_stack` documentation added to it.

Byte figures across six files are re-derived and **dated**. The `exec` row was already stale at 8,367 against a real 9,206, which is why the whole table was re-measured rather than the one changed row patched; the same pass caught `token-budget.md` describing the narrowed surface as "eleven tools" where the test asserts thirteen.

## Review round

Four findings, all four acted on. Each was checked against the code and then **measured** before anything was changed.

**A 32-bit C++ throw was not decoded at all** (P1). The reviewer was right, and it was worse than the finding said: not only did `classify` return `other` for a direct throw, the fail-fast stack scan searched a 32-bit stack for a 64-bit `EXCEPTION_RECORD` and found nothing — so the tool reported no thrown object on the whole 32-bit target class, which this server routes at a dedicated worker on purpose.

Measured by building one program twice and having it print its own record and walk its own graph:

| | 32-bit | 64-bit |
|---|---|---|
| parameters a throw raises | **3** | 4 (the last is the image base) |
| `ThrowInfo`'s links | absolute pointers | RVAs from that base |
| `TypeDescriptor::name` | **+8** | +16 |
| `EXCEPTION_RECORD` size / count / params | **80 / +16 / +20** | 152 / +24 / +32 |

Every one of those makes a mismatched reader come back **empty rather than wrong**, which is exactly why no test caught it. The width is read from the engine per target (`GetActualProcessorType` — measured to report `0x14c` and `0x8664` on the two fixtures) rather than from this build, so the fallback worker and a WoW64 attach are right too. The pointer-versus-RVA row needs no branch: a 32-bit target reports no image base, so taking it as zero makes `base + field` the identity.

**The throw scan ran on faults that cannot have buried one** (P1). It was gated on "no throw in this record" — every access violation, every breakpoint. A C++ record outlives the frames that held it, so a fault deeper than an old `try`/`catch` finds that handled exception's object and reports it as `thrown`. It runs on `abort`'s fail-fast alone now, and the rule is a named function in `fault.rs` rather than a `matches!` in the worker, so it is testable without an engine.

**The summary equated every `abort` with an uncaught C++ exception** (P2). One of its causes, not its meaning. It names the throw only when the scan actually found one now, and otherwise says the record does not say which.

**WIL's fields were read on `>= 3` parameters** (P2) while the comment beside them argued for exactly three. Four or more would have put two arbitrary stack values on the wire named `hresult` and `line`.

### The second fixture, and what mutation testing caught

`docs/samples/cppthrow-fastfail-x86.dmp` (129 KB, same source as the x64 one, scrubbed environment) is the fixture. Forcing the width back to 64-bit reproduces the old behaviour exactly — no `thrown` at all — and the corrected summary then declines to name a cause, which is the two fixes composing.

Every new rule was mutation-verified against the mutation it is for. One result is worth recording: moving `TypeDescriptor::name` to the wrong offset was caught **only** by the test that pins the numbers as literals. The fixture builder and the parser read the same `Bitness` table, so every other test moved with the mutation and stayed green — the exact "test passing on a neighbouring rule" this repo's review skill warns about, found by looking for it rather than by a failure.

### Verification after the round

`cargo test`: **663** unit and **101** smoke, all green; `cargo clippy --all-targets` 0 warnings; `cargo fmt --all --check` clean; markdownlint 0 issues over its four globs. Both fault tier tests run with the fixtures' images present **and** moved aside, each printing a `SKIPPED` line naming the claim it could not check — and the 32-bit test's core assertion, the recovered HRESULT, runs unconditionally in both. The 32-bit worker was rebuilt and restaged after every edit and after the commit.

https://claude.ai/code/session_01Ayv1beKpAVkmDJDLfYqoyf


glslang:
@codex review

glslang:
@codex review

glslang:
@codex review

glslang:
@codex review

glslang:
@codex review

glslang:
@codex review

PR #319, opened 2026-09-14
eval: a fourth local model, the reasoning axis, and the draws that took a claim back

glslang:
Three eval runs, the two bench fixes they found, and one retraction. No `src/` changes — this is
all bench tooling and the write-up.

## The runs

**Fifth (2026-09-13): a fourth local model.** `muse-glimmer:30b-mlx` enters at **14/15**, level
with qwen and gemma, on a 131,072 window rather than 262,144 — so it gets a context group of its
own rather than being asked for a window it cannot serve. **Zero off-surface calls in 426**:
`taught` and `wanted` both 0 in every cell, where the previous run still had `wanted`. Not a
controlled re-run of anything — the server moved 311 commits, the surface 51 → 61 tools, ollama
0.32.15 → 0.34.0, and all three existing tags were re-pulled.

**Sixth (2026-09-14): the reasoning axis, as an A/B.** Two arms differing by one field. Arm A
re-measures the baseline rather than reusing the fifth run, because the fence change below *reaches
these cells* — 24 previously-refused calls now execute. At one draw: **50/60 → 55/60, with fewer
tool calls (143 → 116) and fewer turns, at +70% wall clock.** Reasoning bought *less* tool use, not
more.

## The retraction, which is the point of the last commit

The one-draw write-up called `arm64_pc` "the task that had never been answered, answered". Five
draws of every `min` cell in both arms say otherwise:

| `arm64_pc` on `min` | think off | think on |
| --- | --- | --- |
| muse-glimmer:30b | **2/5** | 5/5 |
| qwen3.8:27b | 1/5 | 3/5 |
| gemma4:31b · nemotron-3.5 | 0/5 | 0/5 |
| **local total** | **3/20** | **8/20** |
| Opus + Sonnet (off) | 6/10 | — |

The same model answers it **2 of 5 times with reasoning off**. The previous run's `0/20` was an
unlucky sample and not a floor; the controls' `2/2` is really 6/10. The fifth run's table keeps its
original claim with a correction beside it — what a run measured is not editable after the fact.

**What survives is better than what was claimed.** Every successful draw calls exactly `open_dump`
then `crash_triage`, as does every failure; neither newly-unfenced tool appears in a single
success. The variance is *interpretation*, and the tool surface is not the variable.

With a denominator: all `min` tasks pooled **60/80 → 68/80, one-sided Fisher p = 0.083**. The
cleanest single effect is gemma decoding the IOCTL with no tool at all, **0/5 → 5/5, p = 0.004** —
and **muse-glimmer going 5/5 → 3/5 on that same task** is the row that keeps it honest. Reasoning
is not uniformly good, and the aggregate hides a real regression.

## The two bench fixes

**A third hex separator** (`0358ef4`). The grader stripped a backtick and an underscore between hex
digits and not an apostrophe, so one Opus answer was graded `n` and `Y` for the same address on two
surfaces. Widened to exactly the three glyphs the evidence names — `-` (23 occurrences) and `,`
(14) are deliberately excluded, being a sample's file name and a decimal separator. Mutation-verified:
exactly two verdicts flip, both control rows, and the three published logs re-grade byte-identical.

**The read-only fence comes from the server now** (`a13b678`). `ALLOWED` was sixteen hand-kept
names and had drifted: the 2026-09-13 run spent 21 calls being refused read-only tools by *this
script*, worded like the server's refusal. It now derives from each tool's `readOnlyHint` (61 of 61
carry it, checked on the wire). Two subtleties in the diff worth a reviewer's eye:

- **an unserved name is handed to the server, not intercepted** — otherwise `off_surface` silently
  becomes `refused` and the bench reports a server that stopped advertising when nothing changed;
- **a served tool with no boolean hint stops the run** rather than being guessed either way.

## Review notes

- **The axis is recorded, not assumed.** `think: true` is a request, like `num_ctx` was when five
  cells asked for 8,192 and were served 32,768. Every turn records `thinking_chars` — 0 across the
  whole `off` arm, 999,210 across the `on` arm's draws. Older logs read `reasoning unrecorded` and
  are deliberately **not** back-filled to `off`.
- **The fence is wider** (37 of 61 on `full`, against 16 before), so tool-choice figures do not
  compare across `a13b678`. Said in the doc.
- **`--compare` reports `harness` and `weights` as moved between the arms**, which is composition
  rather than change: arm B has no Claude rows. Documented rather than silenced.
- `eval-out/` is gitignored; `docs/eval-runs.json` is the committed series, now six runs, with the
  first four verified byte-identical apart from gaining the new `reasoning` key.
- Both CI gates pass on the rebased tree: markdownlint over 48 files, and `cargo fmt --all --check`.

https://claude.ai/code/session_01URVvtdi7p6GeXBiEM2Zbpd

## Summary by CodeRabbit

- **New Features**
  - Added support for comparing model responses with reasoning enabled or disabled.
  - Tool access now follows the server’s current read-only settings across full, lean and minimum tool surfaces.
  - Added evaluation coverage for an additional local model and expanded server configurations.
  - Evaluation results now record reasoning state and measure thinking activity.

- **Documentation**
  - Updated local model evaluation guidance with reasoning comparisons, tool-access behaviour and benchmark findings.
  - Added evaluation plans and recorded results for the 13–14 September 2026 runs.


glslang:
Both findings were real. Fixed in `3e8a31c`, each mutation-verified against the mutation it is for.

**P2 — deny before the surface check.** Confirmed: `wait_for_stop` is in the `exec` group, which neither `lean` nor `min` serves, so naming it there was refused here and recorded `refused_by_harness` instead of reaching the server as `off_surface` — the same corruption the passthrough two lines above exists to prevent. The deny is a statement about tools a client *has*, so it now runs after the membership test. No log on disk moves: `wait_for_stop` has never been called in any run, which is why a green run kept the ordering.

**P1 — the resume key.** Confirmed: two groups differing only by `think` collide and the second is skipped.

Of the two remedies offered, I took the second. Keying cells by `think` would break the experiment the axis was added for: `--compare` pairs cells on `(backend, model, num_ctx, surface)`, so arm A's cells and arm B's would share no key and every row would print `(old)`/`(new)` with nothing compared. The arm is a property of the *run* — like the server build and the weights, which is why it lives in `identity()` — so a plan mixing arms into one log is now refused, naming the colliding coordinate.

**The second half of that finding was right about a hazard, and my first fix measured it wrongly.** Concatenated logs do not "collapse both arms" into an average: `records()` keeps the last record per (cell, draw, task), so the later arm *replaces* the earlier, cell for cell. A concatenated A/B grades **identically to arm B alone** and its identity line reads `reasoning on` — two true sentences about a table that silently dropped half its input. My first version counted arms after the dedup, so it could never fire, and didn't. It reads the raw log now and says what actually happens.

Verified: fires on a concatenated log (12 cells), silent on each arm alone and on logs predating the axis; the plan guard fires on a mixed plan and not on the checked-in ones; all six runs' cells re-grade byte-identical.

glslang:
Round two: three acted on, one declined. All in `f150415`.

**Codex P1 — the guard never inspects the log.** Real, and the same silent-wrong-result class as the first. Copying a plan, flipping `think` and leaving `out` alone is the ordinary way to write the second arm, and a plan-only check cannot see it: every cell reads as finished, the run does nothing and exits 0, and the log still holds the other arm. Since a run that measures zero tasks is indistinguishable from a resume with nothing left to do, it is refused before resuming. Verified: an arm-B plan pointed at arm A's log is rejected by name; both checked-in plans still resume.

**Codex P2 — `--series` and per-log suites.** Real, and worse than stated: `docs/eval-runs.json` spans a task rewording, was built by two invocations merged by hand, and so could not be regenerated by the command the page documents. Each log is now asked which suite it ran against — every record carries it, and a log outlives the plan beside it — with logs predating that field falling back to the argument. **The documented one-liner now reproduces the committed file byte for byte**, which it could not do before.

**CodeRabbit — classify one-sided identity entries as composition.** Right, and better than what I did, which was to document it as a wart on the grounds that the flag erred toward naming too much. `the weights changed` is the one line a reader takes at face value and it was false. The comparison asks only where both runs have an answer now; the two arms report `moved: reasoning`. Mutation-verified in both directions, since a flag that stops crying wolf can as easily have stopped barking: a shared model given a different digest is still named, and so is a harness bumped on two runs that both carry Claude rows.

**CodeRabbit — stale forward reference** to the five-draw run as not yet done. Fixed; it had run by then.

---

**Declined: `think` in the within-log cell key** (Codex round 1, CodeRabbit major). The fact is real and I have not disputed it. The remedy would break the experiment the axis exists for: `--compare` pairs cells on that key, so arm A's cells and arm B's would share none and every row would print `(old)`/`(new)` with nothing compared — measured, not assumed. CodeRabbit's variant keeps a second key notion for pairing alone, which adds a mechanism to a class now closed from both ends: a plan mixing arms is refused, and so is a plan whose arm contradicts its log. That seam has produced four findings; the direction that ends it is fewer mechanisms, not more.

All six runs still re-grade byte-identical; 48 files lint clean.

glslang:
Round three: all three acted on, in `afe0d91`. The first one changed my mind about round one's decline.

**Codex P1 — legacy records claimed by a requested arm.** Real. And with it, **four findings have now landed on one seam**: a plan mixing arms, a plan pointed at the other arm's log, a plan against a pre-axis log. Each of my fixes was a guard bolted onto a resume key that could not tell two arms apart, and the count of guards was going up every round — which is the signal to delete the thing generating them rather than catch the next one.

So `already_done` keys on the arm now, **and only `already_done`.** The cell identity that `summarise`, `matrix` and `compare` share is untouched, because that is what pairs arm A's cells with arm B's. That distinction is why I declined this in round one and it still holds — but the half of the finding that was right is exactly this half, and I should have separated them then rather than declining the whole thing.

Absence reads as `off` for resume and still `unrecorded` for identity, deliberately: a published row must not assert more than its log says, while resume is answering "has this work been done", where the absence is provable — `chat()` sent `think: false` unconditionally until the axis landed. Verified both directions: the pre-axis 2026-09-13 plan still resumes with 38 cells skipped and nothing re-run, and a `think: true` plan against that same log is refused.

**Codex P2 — series exporting a pooled log.** Real, and refusing beats marking. `--grade` warns beside a table someone is looking at; a series row outlives its terminal, and `records()` has already reduced each coordinate to the later arm — so the row would publish one arm's scores under a run identity naming both. There is no honest projection of that log, so the export stops. The six real logs still export byte-identically to the committed file.

**Codex P2 — Fisher on clustered outcomes.** Right, and this one changes a published claim rather than a mechanism. The test treated 80 outcomes per arm as 80 independent trials; they are four tasks × four models × five draws, so each shares a model effect with four others and a task effect with fifteen. That pseudoreplication makes the p-value anti-conservative — it answers a question about independent trials nobody ran. The p-values are gone. The page now reports the rates and the per-model blocks the design actually measured (of four models, two moved and two did not) and says why it quotes no test. The `p = 0.004` on gemma went with them, for the same reason.

All four real plans still resume with nothing re-run; `docs/eval-runs.json` still regenerates byte-identically from the documented one-liner; 48 files lint clean.

glslang:
Round four: both acted on, in `9e2d2b7`.

**`suite_for` rebuilt the recorded basename under `tools/` and ignored the caller's path.** Real. A log records the suite's *basename* — a machine-specific path is the wrong thing to publish in history — and the driver takes an arbitrary tasks file, so a run graded against a suite kept elsewhere came back `FileNotFoundError`, or worse was graded against a same-named stranger under `tools/`. The caller's path now wins when it is the suite the log names, `tools/` is tried next, and anything else is a refusal naming the file it wanted rather than a traceback or a silent substitution. Verified: a log naming a suite outside `tools/` grades correctly when that path is passed, and refuses by name when it is not.

**The fence paragraph said every run on this page predates the derived fence.** Right, and it contradicted the A/B section two screens later — the sixth run ran *behind* that fence, which is the stated reason its baseline arm was re-measured rather than reused. Now scoped to the first five, with the sixth named as the first measured behind it.

Regression after both: the six logs still export byte-identically to the committed `docs/eval-runs.json`, both arm plans still resume with nothing re-run, 48 files lint clean.

---

Running total across four rounds: **12 findings, 11 acted on, 1 declined** (`think` in the shared cell identity — declined on the pairing argument, though round three took the half of it that applied to resume). Four of the twelve landed on the same seam, which is what prompted keying resume by the arm in `afe0d91` rather than adding a fourth guard.

glslang:
Round five: all four acted on, in `e8933f9`. Two were numbers this page had wrong about its own history.

**The allowlist was thirteen, not sixteen.** Checked against the commit the first five runs ran at rather than recalled — and the sixteen-name version turns out never to have existed outside a working tree: it was the thirteen plus three read-only names I added, and the list was deleted outright before anything ran behind it. So every figure derived from it was wrong with it, including `37 of 61 where the list ran 16`, now 13. The `a13b678` commit message says sixteen too; I have left it and put the correction in `e8933f9` rather than rewriting pushed history.

**The identity block does not say `think: false` for the earlier runs** — the page claimed it two paragraphs before explaining why it deliberately does not. Those logs predate the field and read `reasoning unrecorded`. Both halves are now stated together, because the distinction is load-bearing: resume reads that absence as `off` (the driver could not have reasoned), and the identity block refuses to (a published row must not assert more than its source).

**`think` on a `claude-code` group is refused rather than skipped.** `run_cell` does not propagate it to that backend, so those records carry none and `arm_of` reads them as `off`, while the resume lookup keyed from the group would hunt an `on` arm that can never be written — re-running the whole cell every invocation. Keying non-ollama cells as `off` fixes the re-run and keeps the false claim; refusing removes both.

**`suite_for` asked the first record and now asks every one.** Rewording and resuming is documented and legitimate, so a log resumed across a renamed suite holds both names — and for an append-only log the first is the *older*, which would have graded every re-run record against the question it was re-run to escape. With more than one named, the argument decides and the run says so.

Regression: `docs/eval-runs.json` still regenerates byte-identically, all three plans still resume with nothing re-run (48/32/38), the out-of-tree-suite case still grades, 48 files lint clean.

Running total: **16 findings over five rounds, 15 acted on, 1 declined.**

glslang:
Round six: acted on, in `f2c0d22` — and taken as a signal rather than a one-off fix.

`--matrix` (and the `--compare` built on it) did read a pooled log silently. That is **the third reader in three consecutive rounds** — `--grade` warned, `--series` refused, now these — and each of my fixes was the same check bolted onto another caller. Adding a fourth would have kept the count going up.

So the check moved into `records()`, the one function every reader passes through, and the three per-reader mechanisms it replaces are deleted: the `arms` field on a summarised cell, the warning under `--grade`'s table, and the refusal inside `series()`. Four readers, one gate, and a reader added later inherits it instead of having to remember it.

It **refuses** rather than warns, now that it is one decision rather than three. There is no honest reading of such a log — dedup keeps whichever arm was appended later, so every path reports one arm's numbers under an identity naming both, which `--grade`'s warning asked a reader to notice while the table above it stayed just as wrong.

`already_done` is deliberately outside the gate: it reads the file itself, tells the arms apart by key since `afe0d91`, and appending is how such a log gets repaired — refusing to *resume* would block the fix rather than the mistake.

Verified: all four readers refuse with exit 1 from one gate; the six real logs still grade, matrix, compare and export byte-identically; both plans still resume with nothing re-run. The raw scan is cached on (path, mtime, size) since `records()` is called several times over one file per grade.

Running total: **17 findings over six rounds, 16 acted on, 1 declined.** Two of the six rounds have now ended with me deleting mechanisms I had added in earlier ones — the resume key in `afe0d91` and this gate — which is the more useful thing the review has been doing than any individual finding.

PR #353, opened 2026-09-20
A switch that would not resolve is the one incompleteness the report does not count (item 89)

glslang:
`reachable_from_dispatch` named four ways a `NOT REACHABLE` can be short of the graph — a halt, a
work bound, bytes that would not read, and since item 83 the jump-table resolver's own caps — and
computed a fifth without reporting it. `walk_function` records where a path ended at an indirect
jump the tables said nothing about, which is what decides whether resolving is worth an engine
round trip at all; `reachability` read it once for that decision and dropped the final walk's copy.
So a walk that ran the resolver over a dispatch switch and got nothing back carried no signal in
either channel, and the report printed **"Bound hit: no — the reachable call graph was fully
explored"** over a graph missing that switch's every case.

## What landed

- `FnWalk::met_indirect` becomes `FnWalk::unresolved_jumps`, the **sites** rather than a flag, and
  `Report` accumulates them into a `HashSet<u64>`.
- `format_report` counts them, withholds the "fully explored" claim exactly as the blind-instruction
  count already does, and renders a `Jumps not followed` paragraph naming the two remedies that
  reach this — a specific handler VA as `from`, and on a live kernel a module refresh first, a
  table's entries being checked against the image's executable ranges.
- `structured::Reachability::unresolved_jumps` carries the count, skipped when zero beside
  `tables_bounded`. It costs 66 B of output schema and **nothing model-visible**.

The two other silent arms the item named — an instruction set whose operands go unread, and a
listing in no loaded module — need no code of their own: both leave the resolver answering an empty
`JumpTables`, so the walk ends at the jump and the site lands in the same count. One count for three
causes, deliberately: no caller can act on the difference between a table read and not recovered,
one there was nothing to read it with, and one nothing could ask about.

## Two things the item did not see

**The probe.** `reachability` walks each function once with no tables to decide whether resolving is
worth a round trip, and a walk that reaches the target on that pass returns straight away — so its
list is every indirect jump on the way, none of which was ever offered to a resolver. Merged, every
`REACHABLE` found inside a dispatch routine would have reported its own switch as one the walk could
not follow: the ordinary success on the ordinary target, carrying the one signal that says a graph
has holes in it. That early return contributes nothing, which is what lets the rendering sit outside
the verdict branches — a `REACHABLE` found in a partial graph says so for the same reason
`tables_bounded` does, one review round of #351 after it shipped on one side only.

**"Switch" was the wrong word, and only a real target said so.** Driven against the ARM64 kernel
dump on the build under test (`0.18.0+ga8e42816`), ordinary `nt` routines come back with counts that
are mostly *not* switches — 11 on `nt!ObpLookupObjectName` and on `nt!KiDispatchException` inside 24
explored functions, 3 on `nt!NtQuerySystemInformation`, 2 on `nt!NtSetSystemInformation` — because a
tail call through a register is an indirect jump too. The paragraph says *whatever they reach: a
switch's case blocks, or a callee a tail jump goes to*; the first draft called all of them switches
and would have described most of them wrongly.

## Counted per site

`visited` is keyed by the *start* address, so a routine entered at two boundaries is walked twice and
the two walks overlap. Summing what each ended at reports one jump as two.

## Testing

On the ARM64 bench, at `8c4a53b`:

- `cargo test` with `WINDBG_MCP_SMOKE_DUMP=1`: **985** unit tests and **116** `mcp_smoke` tests pass
  (63.4s, so the debugger tier ran rather than skipping). `cargo clippy --all-targets` clean;
  `cargo fmt --all --check` and the markdownlint globs clean on the Mac.
- The surface golden was re-recorded: `reachable_from_dispatch` wire 7306 → 7372, output schema
  4557 → 4623, model-visible unchanged at 91790 B.
- Mutation-verified, four ways, each caught by the test that states the property:
  | mutation | what fails |
  |---|---|
  | the final walk's copy thrown away (the pre-fix behaviour) | all three new tests |
  | the probe's list merged in the resolver arm | the two count assertions (1 → 2) |
  | counted per walk rather than per site | `one_switch_reached_from_two_starts_is_counted_once` |
  | the probe merged on the proven-path return | `a_verdict_proven_without_tables_counts_no_unresolved_switch` |

`FOLLOWUPS.md` item 89, moved to `DONE.md`.

https://claude.ai/code/session_01VtbjUoCGL4aJSQNS96DKLt

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `reachable_from_dispatch` reporting when indirect jumps cannot be followed.
  - Reports now count unresolved jump sites without double-counting overlapping paths.
  - Full-exploration claims are withheld when unresolved jumps remain.
  - Human-readable reports identify jumps that were not followed and provide remediation guidance.
  - Structured results include an `unresolved_jumps` count when applicable.
  - Resolved and probe-only jumps are excluded from the count.

- **Documentation**
  - Updated limitations, structured-results, changelog, and follow-up documentation to explain the new reporting behaviour.


glslang:
@codex review

